PR #1, opened 2019-01-07
Fix clippy lints

fredmorcos:
I did the obvious ones, a couple of long methods are remaining which I'll work on next.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,7 +1,10 @@
+#![warn(clippy::all)]
+
 use std::io::{self, stdin, Write};
 use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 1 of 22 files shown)


Canop:
Thanks for that.

As I had made other changes and as I doesn't want to replace *all* the matchs by if/else, I cherry picked what I found interesting in your PR.

Next time you want to make a big change please come on [the chat](https://miaou.dystroy.org/3) for discussion and coordination.

Canop:
I don't always prefer it.
I probably use more if/else than matchs for booleans.
But there are cases where it feels more consistent with the surrounding code or just more readable.

fredmorcos:
Alright, thanks.

Issue #20, opened 2019-02-04
Use EDITOR and PAGER environment variables to edit and view respectively


zummenix:
Currently, `broot` creates a config file with `edit` and `view` tools predefined. I think, for most users the default/expected behaviour would be to use tools set in `EDITOR` (with fallback to `vi`?) and `PAGER` (with fallback to `less`?) environment variables.

Canop:
The problem is not so many users have those variables defined, which means you need to ask users to define an editor in the config anyway. I didn't see any obvious way to make that clear for users.

Even xdg-open is rarely very relevant.

PR #45, opened 2019-03-12
Allow environment variables in verbs

lovasoa:
This PR implements a new feature that I find useful: environment variables in verbs.
This allows verbs to reference environment variables.

Instead of hard coding `/usr/bin/nvim {file}`, one can now write `$EDITOR {file}`. It will open the file in the default editor.

See #20


lovasoa:
@Canop : What do you think ?

Canop:
I'm not 100% sure it's the right approach but I'll probably merge it as it seems innocuous.

lovasoa:
Yes, allowing reference to environment variables in verbs allows better configurability, but it still does not fully solve #20, as it does not provide a fallback editor.

lovasoa:
Anyway, it should be more widely compatible than the current hard reference to `/usr/bin/nvim`

PR #47, opened 2019-03-12
Avoid code duplication in status.rs

lovasoa:
Remove some code duplication in Screen::write_status_err


Canop:
Please try to avoid such minor refactorings. I'm not sure I prefer the code now and it took us too much time for such a minor concern.

lovasoa:
I do this for fun. Do not hesitate to refuse pull requests if you don't like them.

Issue #27, opened 2019-02-13
Feature request: add key for "select previous"


magnetophon:
With tab you can go to the next match.
I expected shift-tab to go to the previous one.

Canop:
Sounds logical. Will do.

dgrisham:
@Canop Hm, doesn't seem like you can -- see https://github.com/redox-os/termion/blob/master/src/event.rs#L54. Might require a feature request/PR.

dgrisham:
Asked about it here https://gitlab.redox-os.org/redox-os/termion/issues/16#note_15760

PR #52, opened 2019-04-20
Add C-n for next and C-p for previous selection

arnej:
This adds additional key bindings for selecting the next and previous lines. The used key bindings are the same as in e.g. [https://github.com/junegunn/fzf](url) and feel vim-like for me.

I also found [https://github.com/Canop/broot/issues/27](url) which requests a key binding to navigate to the previous line by pressing Shift+Tab, but I couldn't get any output from termion when pressing that key combination.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -140,10 +140,10 @@ impl Command {
             Key::Ctrl('q') => {
                 self.action = Action::Quit;
             }
-            Key::Up => {
+            Key::Up | Key::Ctrl('p') => {
                 self.action = Action::MoveSelection(-1);
             }
-            Key::Down => {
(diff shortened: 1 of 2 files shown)


Canop:
I'd welcome opinions and arguments on this particular choice of shorrcuts.

dgrisham:
I think this is a good default because `n` and `p` stand for 'next' and 'previous`. However, I think `tab` and `shift+tab` are much more intuitive/expected for the average user (I tried those almost immediately after my first broot search). I think `tab` and `shift+tab` are better defaults, and configurable keybindings would be a way for anyone with a preference for `C-n` and `C-p`, or anything else, to switch.

Canop:
Made obsolete by #63

PR #63, opened 2019-07-22
Add support for using backtab to select the previous match

arnej:
This PR adds the ability to use backtab (`Shift+Tab`) to select the previous match as requested in [Issue 27](https://github.com/Canop/broot/issues/27).
I tried to implement this a while ago, but couldn't find a way to get the key combination working. I then created a minimal test case for this key combination using crossterm and it still didn't work.
Then I tried another terminal emulator and finally got it working.

To conclude:

- This didn't work in my testing with konsole (may be related to my settings, but I couldn't make the key combination `Shift+Tab` work in konsole). Maybe someone else can test this change with konsole to see if it works with their settings.

- This works in my current terminal emulator alacritty without any further setup required.

This PR also potentially obsoletes [#52 ](https://github.com/Canop/broot/pull/52), as I created that one to have an option for selecting the previous match without arrow keys (they are far away :)) while I was trying to find a way to get backtab working.

### DIFF
--- a/src/browser_states.rs
+++ b/src/browser_states.rs
@@ -244,6 +244,13 @@ impl AppState for BrowserState {
                     tree.make_selection_visible(page_height);
                 }
                 Ok(AppStateCmdResult::Keep)
+            },
+            Action::Previous => {
+                if let Some(tree) = &mut self.filtered_tree {
+                    tree.try_select_previous_match();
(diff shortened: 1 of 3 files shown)


Canop:
It looks OK. It works on my consoles too.

PR #109, opened 2020-01-10
Update Readme

FabulousCupcake:
* Adds table of contents for increased readability
* Adds homebrew shield


Canop:
@FabulousCupcake thanks for that but I have a problem with the brew shield: how am I supposed to ensure it doesn't contain any trojan as I'm not the maintainer of the brew formula (which I don't want to do as I don't have the competences) ?
I'd like to discuss this point with people familiar with brew ( miaou room: https://miaou.dystroy.org/3 )

PR #118, opened 2020-01-10
Use local variables in br shell function.

ghuls:
Improve br shell function:
  - use local variable to prevent accidental overwriting variables
    set by the user in the current shell
  - rename variables to a more descriptive name
  - use $status instead of undefined $code in fish br function

### DIFF
--- a/src/shell_install/bash.rs
+++ b/src/shell_install/bash.rs
@@ -44,23 +44,23 @@ const BASH_FUNC: &str = r#"
 # It's needed because some shell commands, like `cd`,
 # have no useful effect if executed in a subshell.
 function br {
-    f=$(mktemp)
+    local command_file=$(mktemp)
     (
 	set +e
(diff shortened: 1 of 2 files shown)


Lucretiel:
In fish this won't do what you expect, because it will return the `$status` of the prior `rm` command. #128 has a corrected fish implementation.

Canop:
closed in favor of PR #128

@ghuls a confirmation that all your concerns are fixed now would be welcome

ghuls:
@Canop The bash function is not fixed yet. I can rebase the patch on the last git version.

If you don't have an objection to it, I could like to change `code` to `error_code` in the fish version too.

Canop:
If you rebase I'll study it (and submit it to other readers)

mgritz:
Got something very similar to ghuls' changes so at least his bash changes (in use with zsh). :+1:

PR #128, opened 2020-01-10
Made the fish function implementation more idiomatic

Lucretiel:
This pull request makes the fish function more idiomatic. In addition to a minor refactor of the function itself, to use more descriptive variable names and local variables, it also moves the function definition into `~/.config/fish/functions/br.fish`, where fish stores & lazily loads functions from.

Additionally, `rustfmt fish.rs`


Canop:
There should be a discussion with fish users on this PR.

Lucretiel:
Totally. For reference, the relevant fish documentation for `~/.config/fish/functions` is here https://fishshell.com/docs/current/tutorial.html#tut_autoload

offbyone:
This is what I expect for fish shell integration. +1 from me.

PR #132, opened 2020-01-11
Numerous minor refactorings

Lucretiel:
- Fixed many unidomatic uses of "if let Some(x) = thing"
- Added standard trait implementations and other interface refactorings (for instance, moving "from(thing)" to From<Thing>. All of these do not affect the use of these interfaces.
- Replaced &Box<dyn Trait> with &dyn Trait, where relevant.
- App::run and App::end now take App by value, to enforce that they should only be called once.
- Replaced `loop` with `for event in rx_events` in App::run.
- Touched files were rustfmt'd

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -9,18 +9,9 @@
 //! - a request to launch an executable (thus leaving broot)
 
 use {
-    std::io::Write,
-    crossterm::{
-        cursor,
-        event::{DisableMouseCapture, EnableMouseCapture},
-        terminal::{EnterAlternateScreen, LeaveAlternateScreen},
-        QueueableCommand,
-    },
-    minimad::Composite,
-    termimad::EventSource,
     crate::{
         app_context::AppContext,
-        app_state::{AppStateCmdResult, AppState},
+        app_state::{AppState, AppStateCmdResult},
         browser_states::BrowserState,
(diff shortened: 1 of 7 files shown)


src/app.rs
```diff
@@ -185,55 +181,53 @@ impl App {
...
+        let state = self.mut_state();
+        state.display(writer, &screen, con)?;
+        state.write_status(writer, &cmd, &screen, con)?;
+        state.write_flags(writer, &mut screen, con)?;
+
         screen.input_field.display_on(writer)?;
 
-        loop {
-            let tl = TaskLifetime::new(event_source.shared_event_count());
-            if !self.quitting {
```
Lucretiel: Correctness note: the only way that self.quitting can be true is through a call to `apply_command`, which means that moving it to the end of the loop is acceptable here

Canop:
Thanks.

Most of this is good improvement/cleaning and I agree with.

Next time, please come to my chat so that we can discuss the changes (I'm for example a little unclear at the interest of allocating only once the atomicusize and I wouldn't have changed the event loop which is very linked to how I implement and refactor the event issuer in termimad).

Lucretiel:
Sure thing, thanks for the heads up!

PR #133, opened 2020-01-11
Various string-related enhancements

Lucretiel:
- Replaced a few cases of `slice::join` and `format!` with non-allocating versions
- Removed a few unnecessary uses of to_string


Canop:
I check the crates I import, and I want to keep them to the minimum. I see no reason for those changes.

PR #149, opened 2020-01-13
website: add information on installing broot via MacPorts

herbygillot:
broot is now present in MacPorts; adding instruction for installing it that way to the website docs.


Canop:
It looks fine.

herbygillot:
Thank you @Canop

PR #166, opened 2020-01-18
Allow printing relative paths as well

roshan:
This is a use-case I have where I want to sometimes reference things relatively. This lets me do `file:prp` and have it printed relative to the CWD.

This is just the code I use. Happy to rewrite it to your standards to upstream.


roshan:
Thank you!

PR #213, opened 2020-03-14
Add Skins section to nav and solarized dark skin

danieltrautmann:
The navigation now has a new section called Skins and the Solarized Dark skin was added.


Canop:
@danieltrautmann Skins have changed in broot, most notably to allow differentiation of non focused panels. See https://dystroy.org/broot/skins/ . Are you willing to update your skin ?

danieltrautmann:
@Canop Yes, I will update the skin. Thanks for notifying me about that. I might find some time during the weekend.

PR #272, opened 2020-08-31
Draft: Clipboard support, copy to clipboard

asdf8dfafjk:
Draft: Support copying current path to system clipboard.

Test here: https://github.com/asdf8dfafjk/broot_testing/commit/ba80534f556a5d3b3ae64bfdfee2b93b4ade140a


asdf8dfafjk:
This is a draft for supporting copying a path. I still need to add installation instructions for `xcb` .

This adds an additional library dependency for `xcb` both at compile and run-time. There are no particular downsides to a compile time dependency (unless you'd like to keep it dependency free). There is a major downside to compile time dependency that the program will absolutely not run (rather than gracefully not having a feature). There are apparently better ways to do this but I have no experience there and might not afford to spend time.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ client-server = []
 ansi_colours = "1.0"
 bet = "0.3.4"
 clap = { version="2.33", default-features=false, features=["suggestions"] }
+cli-clipboard = {path="/home/zx/workspace/git/cli-clipboard"}
```
asdf8dfafjk, line 24: I looked at a lot of crates and this was simplest in that it didn't follow "rust dogmatism", no need to create a clipboard instance (because system clipboards are as global as they come) etc.

src/app/state.rs
```diff
@@ -81,6 +81,13 @@ pub trait AppState {
         let con = &cc.con;
         Ok(match internal_exec.internal {
             Internal::back => AppStateCmdResult::PopState,
+            Internal::copy_path => {
+                cli_clipboard::set_contents( self.selected_path().to_string_lossy().into_owned() )
+					.map_err( |_| ProgramError::ClipboardError )?
+				;
+
+				AppStateCmdResult::Keep
+            }
```
asdf8dfafjk, line 90: I'm not entirely sure if `AppStateCmdResult::Keep` is the best way. I looked at and liked the behavior or `line_down` where one just has to press enter. One can argue that if I need to copy multiple paths, I'd like to keep the command as is, use arrows to navigate, and hit enter every time I want to copy. 

I initially wanted to clear the command prompt upon invoking this command but couldn't figure how

src/browser/browser_state.rs
```diff
@@ -328,6 +328,14 @@ impl AppState for BrowserState {
...
             }
+            Internal::copy_path => {
+                let path = &self.displayed_tree().selected_line().target();
+                cli_clipboard::set_contents( path.to_string_lossy().into_owned() )
+					.map_err( |_| ProgramError::ClipboardError )?
+				;
+
+				AppStateCmdResult::Keep
+            }
             Internal::focus => internal_focus::on_internal(
```
asdf8dfafjk, line 339: Same, `AppStateCmdResult::Keep` best choice?

src/errors.rs
```diff
@@ -11,6 +11,7 @@ custom_error! {pub ProgramError
     Io {source: io::Error} = "IO Error : {:?}",
     Crossterm {source: crossterm::ErrorKind} = "Crossterm Error : {:?}",
     Termimad {source: termimad::Error} = "Termimad Error : {:?}",
+	ClipboardError = "Clipboard handler error: unable to copy to clipboard",
```
asdf8dfafjk, line 14: I decided to not include the actual error, because I didn't want to reimport `anyhow::Error` , and I don't expect clipboard actions to fail ...

src/verb/builtin.rs
```diff
@@ -33,15 +33,17 @@ pub fn builtin_verbs() -> Vec<Verb> {
             "/bin/cp -r {file} {newpath:path-from-parent}",
             StayInBroot,
         )
-        .unwrap()
```
asdf8dfafjk, line 36: My default keybindings are Tab, but I replace them with spaces, I intended to but I forgot

website/docs/install-br.md
```diff
@@ -39,7 +39,7 @@ You can bind this command to a key sequence in the [configuration file](../conf_
 
 ```toml
 [[verbs]]
-key = "alt-c"
+key = "alt-p"
```
asdf8dfafjk, line 42: I know this is just an example but I didn't want to mislead people.

asdf8dfafjk (commented):
Let me know what you think....

Cargo.toml
Canop: Why did you link to a local copy ? I changed that to `cli-clipboard = "0.2"` and it looks fine

Canop:
Thanks. It looks very good.

If I manage to find how to compile it for non unix platforms (right now it fails for me at least for windows), I'll make a few fixes and use it as the basis for a new branch (so that Mac and Windows users can test too).

Canop:
I've merged this into a new "clipboard" branch. We'd need now to solve the cross-compilation problem :\

Cargo.toml
asdf8dfafjk: Sorry I left that in by mistake.

PR #279, opened 2020-09-16
Support wayland.

asdf8dfafjk:
Use alacritty's copypasta.

Compilation: 

1) Tested with x11 linux, `cargo build --release --features clipboard` compiles and works as expected
2) Tested on osx using "fastmac". `cargo build --release --features clipboard` compiles just fine! Throws an error probably due to clipboard not being accessible during shell session.

### DIFF
--- a/src/app/state.rs
+++ b/src/app/state.rs
@@ -1,6 +1,7 @@
 use {
     super::*,
     crate::{
+		clipboard,
         command::*,
         display::{Screen, W},
         errors::ProgramError,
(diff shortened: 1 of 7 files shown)


src/app/state.rs
```diff
@@ -91,10 +92,10 @@ pub trait AppState {
                 #[cfg(feature="clipboard")]
                 {
                     let path = self.selected_path().to_string_lossy().to_string();
-                    match terminal_clipboard::set_string(path) {
+                    match clipboard::set_string(path) {
                         Ok(()) => AppStateCmdResult::Keep,
-                        Err(_) => AppStateCmdResult::DisplayError(
-                            "Clipboard error while copying path".to_string()
+                        Err( e ) => AppStateCmdResult::DisplayError(
+							format!( "{}", e )
```
asdf8dfafjk, line 98: This prints an error that's "too techinical" I think, "Clipboard Error #NS<somethign somethign> failed to initialize". I am in favor of just saying: "Clipboard error" personally

Canop:
Are there really cases where this allows some clipboard operations that aren't already working ?

asdf8dfafjk:
I don't remember clearly now but when I submitted this, I had a reason to believe that it didn't work on Wayland and machinetosh. 

Since you are not interested in incorporating smithay, I figure the point is moot.

PR #296, opened 2020-10-28
feat: instant break for circular symlinks

drahnr:
Tiny improvements to avoid going in circles longer than need or not getting to the end of a long link chain due to the heuristic upper limit is kicking in.


Canop:
Seems perfect, thanks.
I'll test later today.

PR #300, opened 2020-11-08
Set window title

asdf8dfafjk:
Kludgy but apparently only way 

https://stackoverflow.com/questions/23388262/is-there-a-way-to-change-the-terminal-title-with-curses

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -41,6 +41,7 @@ fn configure_log() {
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


Canop:
Sorry but I don't want this. I don't want to do too much and interfere with the normal behavior of terminals and user customization.

This could be done in an independent wrapping or a personnalisation of a launch script.

asdf8dfafjk:
Np!

PR #307, opened 2020-11-24
fix link to vscode.ttf

SiegfriedEhret:
Hello!
In https://dystroy.org/broot/icons/ the link to the ttf file ends in a 404 error.
Here is a fix using the master branch version, I hope that's ok.
Peace!


Canop:
Thanks for the fix!

PR #311, opened 2020-12-09
Docs: Vim integration plugin

lstwn:
Hey,

first of all thanks for making this awesome piece of software! 

Since I found no integration plugin for vim, I created a small one. To help people looking for vim integration as well, I added a reference to the plugin in the tips and tricks section of the docs on the website.

PS: Super sweet that one can join multiple configuration files for broot via the command line flag. However, this info was hidden in the help section and not mentioned elsewhere. However, it is massively useful for integrations like this (I used that feature to append "enter on a file prints the path behavior" to the default configuration). It might be worth pointing that out in the launch section of the docs (next to --out and the likes). Let me know, if I can help..

### DIFF
--- a/website/docs/tricks.md
+++ b/website/docs/tricks.md
@@ -121,3 +121,8 @@ Broot usage, just like vim, relies a lot on the <kbd>esc</kbd> key. If you're a 
 
 This brings a lot of comfort, not just in broot.
 
+## Vim integration/plugin
+
+In case you want to use broot for opening files fuzzily in vim (and potentially
+replace netrw), check out: [broot.vim](https://gitlab.com/lstwn/broot.vim)
+


vanhtuan0409:
@lstwn can we consider adding support for nvim also :D I'm desperate for this plugin

lstwn:
> @lstwn can we consider adding support for nvim also :D I'm desperate for this plugin

I wouldn't mind accepting a PR for nvim compatibility. Since the plugin relies heavily on vim8's terminal feature (which isn't compatible with nvim's terminal feature as far as I know), it is not just a simple change of commands to make it compatible I think. Also, since I'm not using nvim, I cannot devote my time atm.. Feel free to fork it and create a PR though!

PR #312, opened 2020-12-09
FIX: faulty default config

lstwn:
This fixes an issue with the default config: If you were to uncomment the `[special-paths]` table/dictionary and the `date_time_format` in the default config, you are greeted with an error `Unrecognized: YmdR` because the `date_time_format` is then parsed into the special paths table because TOML.

Basically, what the comment in the beginning warns people about.

The new order fixes that issue.


Canop:
Thanks!

PR #315, opened 2020-12-13
add Alpine Linux into third party repositories

jirutka:
Note: Alpine 3.13 has not been released yet, it should be till the end of this year.


Canop:
Thanks

Issue #337, opened 2021-02-01
Fish shell function: use `--wraps`


basbebe:
This way, `br` can use the shell completions installed for `broot`.

```fish
[…]

function br --wraps=broot
    […]
```

https://fishshell.com/docs/current/cmds/function.html
> `-w WRAPPED_COMMAND` or `--wraps=WRAPPED_COMMAND` causes the function to inherit completions from the given wrapped command.

Canop:
Can you test the whole and make a PR ?

PR #338, opened 2021-02-01
fish completion for br

basbebe:
make `br` inherit completion of `broot`

https://fishshell.com/docs/current/cmds/function.html
> `-w WRAPPED_COMMAND` or `--wraps=WRAPPED_COMMAND` causes the function to inherit completions from the given wrapped command.

https://github.com/Canop/broot/issues/337


Canop:
So that I know if I must do it myself: did you test the shell function installation process with this and the subsequent completions ?

basbebe:
I did test the shell function.
I did not test if the change changes anything in the build or breaks installation…

PR #355, opened 2021-03-02
add vim-floaterm broot wrapper link

voldikss:
That plugin supports both nvim and vim.

### DIFF
--- a/website/docs/tricks.md
+++ b/website/docs/tricks.md
@@ -188,3 +188,5 @@ This brings a lot of comfort, not just in broot.
 In case you want to use broot for opening files fuzzily in vim (and potentially
 replace netrw), check out: [broot.vim](https://gitlab.com/lstwn/broot.vim)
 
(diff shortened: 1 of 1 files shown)


Canop:
Interesting.

I'll have to have a deeper look at floaterm before I link to it: there seems to be a problem in [the plugin](https://github.com/voldikss/vim-floaterm/blob/master/autoload/floaterm/wrapper/broot.vim), maybe a problem of compatibility with last broot, it doesn't work correctly on my computer (b&w broot, no selection shown).

voldikss:
Thank you for the attention!

voldikss:
> maybe a problem of compatibility with last broot

Is that the incompatibily of config file (hjson v.s. toml)?

Canop:
No, hjson or toml aren't a problem, maybe the way removing colors is decided. It's possible a `--color auto` would be needed. But I should fork your project to try test to be sure.

PR #359, opened 2021-03-12
Fix cargo test on windows, make the test OS agnostic

Stargateur:


Canop:
I need to understand precisely what happens here. Can you show me the failure you get on Windows with the string version ?

Stargateur:
```none
running 11 tests
test keys::key_parsing_tests::check_key_description ... ok
test keys::key_parsing_tests::check_key_parsing ... ok
test pattern::fuzzy_pattern::fuzzy_pattern_tests::check_orderings ... ok
test pattern::fuzzy_pattern::fuzzy_pattern_tests::check_match_pos ... ok
test path::normalize::path_normalize_tests::test_path_normalization ... FAILED
test content_search::needle::content_search_tests::test_found ... ok
test verb::verb_invocation::verb_invocation_tests::check_verb_invocation_parsing_empty_arg ... ok
test verb::verb_invocation::verb_invocation_tests::check_verb_invocation_parsing_post_bang ... ok
test verb::verb_invocation::verb_invocation_tests::check_verb_invocation_parsing_oddities ... ok
test verb::verb_invocation::verb_invocation_tests::check_verb_invocation_parsing_empty_verb ... ok
test verb::execution_builder::execution_builder_test::test_build_execution ... ok

failures:

---- path::normalize::path_normalize_tests::test_path_normalization stdout ----
-----------------
normalizing "/abc/test/../thing.png"
thread 'path::normalize::path_normalize_tests::test_path_normalization' panicked at 'assertion failed: `(left == right)`
  left: `"\\abc\\thing.png"`,
 right: `"/abc/thing.png"`', src\path\normalize.rs:41:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    path::normalize::path_normalize_tests::test_path_normalization

test result: FAILED. 10 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Have you an other idea to fix it ?

Canop:
Besides the tests, does the related feature even work on Windows ?

It resolves the ../.. you type in commands:
image

Does that work ? Could you show me what's displayed in such a case and confirm it works on enter ?

Stargateur:
image

I guess it's work the same

Stargateur:
I did some test, problem is that this doesn't test anymore your "end with slash" feature. I don't think it's possible to make a test that check that, your hack:

```rust
if ends_with_slash {
    normalized.push("");
}
````

doesn't look very robust, I don't think this behavior is mandatory since there is no difference from `/my_file` and `/my_file/` from OS point of view. If you use [`components()`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.components) you will never see a empty component:

```none
---- path::normalize::path_normalize_tests::test_path_normalization stdout ----
-----------------
normalizing "/abc/test/../thing.png"
RootDir
Normal("abc")
Normal("thing.png")
-----------------
normalizing "/abc/def/../../thing.png"
RootDir
Normal("thing.png")
-----------------
normalizing "/home/dys/test"
RootDir
Normal("home")
Normal("dys")
Normal("test")
-----------------
normalizing "/home/dys"
RootDir
Normal("home")
Normal("dys")
-----------------
normalizing "/home/dys/"
RootDir
Normal("home")
Normal("dys")
-----------------
normalizing "/home/dys/.."
RootDir
Normal("home")
-----------------
normalizing "/home/dys/../"
RootDir
Normal("home")
-----------------
normalizing "/.."
RootDir
ParentDir
-----------------
normalizing "../test"
ParentDir
Normal("test")
-----------------
normalizing "/home/dys/../../../test"
RootDir
ParentDir
Normal("test")
-----------------
normalizing "π/2"
Normal("π")
Normal("2")
-----------------
normalizing "/home/dys/dev/broot/../../../canop/test"
RootDir
Normal("home")
Normal("canop")
Normal("test")
```

I still didn't really understand why you want to keep the slash information. Anyway, I think you should create a wrapper that will really give strong felling about the end slash:

```rust
struct PathBufSlash {
    inner: PathBuf,
    ends_with_slash: bool,
}

pub fn normalize_path<P: AsRef<Path>>(path: P) -> Result<PathBufSlash> {
    let ends_with_slash = path.as_ref().to_str().map_or(false, |s| s.ends_with('/'));
    let mut normalized = path.canonicalize()?;

    Ok(PathBufSlash { inner: normalized, ends_with_slash })
}
```

I think this make your code way more robust.

Canop:
> doesn't look very robust, I don't think this behavior is mandatory since there is no difference from /my_file and /my_file/ from OS point of view. 

It's not for OS at all. I don't care about the OS. It's for user interaction. If he user typed a / at the ends, it means it's a directory and the next tab must go inside, not propose name completions.

Stargateur:
> 
> 
> > doesn't look very robust, I don't think this behavior is mandatory since there is no difference from /my_file and /my_file/ from OS point of view.
> 
> It's not for OS at all. I don't care about the OS. It's for user interaction. If he user typed a / at the ends, it means it's a directory and the next tab must go inside, not propose name completions.

What I mean is the nothing in PathBuf in std guaranty this behavior.

Canop:
Besides the test, which could be fixed with a cfg/feature thing, a annoying limit is that the `normalize_path` code tests whether it ends with a `/` and might lead to sub-optimal interaction on Windows.

I'll try to grab a windows computer and do some tests when I can.

Stargateur:
There is https://doc.rust-lang.org/std/path/fn.is_separator.html I think that the way to go

PR #360, opened 2021-03-17
Implemment a wrapper for PathBuf that do its best to keep the endind separator

Stargateur:
This is a proof of concept about my idea in https://github.com/Canop/broot/pull/359#issuecomment-799729867. There is a lot to be improved, I think as you have specific need you should use two wrappers for `Path` (maybe avoid allocation ?) and `PathBuf`.

Also I think instead of building my `PathBufWrapper` from a `PathBuf` (as this PR currently do), broot should build it from the raw str.

I keep `normalize_path()` cause https://doc.rust-lang.org/std/fs/fn.canonicalize.html change the behavior, I don't have a full understanding of broot internal to know if we can use the std one and it's return a possible error.

This handle every OS separator using https://doc.rust-lang.org/std/path/fn.is_separator.html.


Canop:
PathBuf clearly lacks some essential features and could be replaced but be careful that:

* I'm usually not willing to make my code more complex and can have a disappointing answer to pull requests
* everything in broot is constrained by performances needs

Stargateur:
Currently, you code is more complex cause PathBuf is not build to have this feature and so you expect using a hack that it will work. My proposition is a code simplification.

In term of performances, I think this is strictly equivalent, but I think we can improve it with a PathWrapper.

PR #362, opened 2021-03-19
add colored help

wfxr:
Thank you for maintaining this awesome tool in advance.

Lots of cli tools written in rust have a colored help (fd, bat etc.). I think adding this tiny feature to `broot` is not a bad idea :)

image

( I noticed that `default-features` for clap are disabled in the `Cargo.toml`. Since the `color` feature is one of them I don't know if it was intentional disabled. )

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ ansi_colours = "1.0"
 base64 = "0.13"
 bet = "0.4.0"
 char_reader = "0.1"
-clap = { version="2.33", default-features=false, features=["suggestions"] }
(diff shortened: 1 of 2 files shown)


Canop:
It was intentional because I didn't really like the coloured clap output but I'll try again with your PR.

wfxr:
@Canop Got it. Thanks for your explanation. I respect your choice. Feel free to close this if you still don't like it

Canop:
It might be the right time for people to say whether they find the colored output prettier. For now I don't have anything strong against it.

Canop:
We'll revert if people complain it's hard to read on their terminals (I know it's not very good with some light terminal settings).

PR #390, opened 2021-06-01
add solarized theme to doc.

guillaumecherel:


Canop:
It will be live in dystroy.org/broot when I deploy the 1.5 version of broot, tonight or in the next few days

PR #417, opened 2021-07-05
Document the --color option

Wizhi:
I've simply added the `--color` option to the `man` page, whilst removing `--no-style`, which was seemingly removed sometime between 93b1ffcc701a7753e8376c80b0c10683f7e3e36d and 7d512894808500946a6acb6d4939d4ba1473c14d.


Canop:
Hum... this file was supposed to be generated... what did I break ?

Canop:
I just had a second look. Man page generation isn't yet automated. This is unfortunate, I'll add an issue for that.

Thanks @Wizhi  for the update

PR #449, opened 2021-09-28
Add Gentoo Linux installation instructions

creideiki:
Broot is available in Gentoo Linux: https://packages.gentoo.org/packages/app-misc/broot

Add it to the list of distro installation instructions.


Canop:
It's online: https://dystroy.org/broot/install/#gentoo-linux
Thanks

PR #526, opened 2022-03-28
feat: replaced all the name records in the icon font from `Untitled1` to `Broot Icons Visual Studio Code`

stevenxxiu:
I used *FontForge* to do this. Replaced all `Untitled1` to `Broot Icons Visual Studio Code` in `vscode.ttx`.

Copy of https://github.com/Canop/broot/pull/525 with a better branch name.


Canop:
Anybody with some slight knowledge about fonts to check this ?

Issue #535, opened 2022-04-15
Program not found on windows


Spacelord-XaN:
OS: Windows 10 Version 21H2
Shell: PowerShell 7

The following commands do not work:

- mkdir
- cp
- mv
- rm

For example the command mv results in these log entries:
```
08:59:14.569 [INFO] broot::verb::external_execution: Executing not leaving, launchable Program { exe: "mv", args: ["C:\\temp\\test.csv", "C:\\temp/test2.csv"], working_dir: None, capture_mouse: true }
08:59:14.570 [DEBUG] broot::launchable: working_dir: None
08:59:14.581 [WARN] broot::verb::external_execution: launchable failed : LaunchError { program: "mv", source: Error { kind: NotFound, message: "program not found" } }
```

Canop:
Can you try to define verbs with complete paths to the programs ?

Something like

```
{
    invocation: "md {subpath}"
    external: ["c:/path/to/mkdir.exe", "-p", "{directory}/{subpath}"]
}

```

Stargateur:
I think the correct default should be to use `Windows` command https://www.lemoda.net/windows/windows2unix/windows2unix.html for powershell at least.

Spacelord-XaN:
@Canop 
Yes that works fine

I think the main problem is, that under Windows there is no such thing as mkdir.exe.
As far as I understand it in PowerShell mkdir is just an alias for the the New-Item function and in the classic CMD box mkdir is a built in function of the CMD box.

Two possible workarounds come to my mind:
With git for Windows there comes a bunch of .exe files (like mkdir.exe, cp.exe, mv.exe) located under:
`c:\Program Files\Git\usr\bin\`
In there one can find a mkdir.exe, and if you add this directory to your path then this might work (I have to test this)

Another way would be to start the cmd.exe and pass the command to execute to it.
For example:
`cmd.exe /c mkdir test123`

And as a verb it would look like this:
```
{
    invocation: "md {subpath}"
    external: "cmd /c mkdir {subpath}"
    leave_broot: false
}
```

Canop:
We could define all built-in internals like ` ["cmd", "/c", "mkdir", "{subpath}"]` on Windows, for example with `#[cfg]` attributes in [there](https://github.com/Canop/broot/blob/master/src/verb/builtin.rs#L48).

But I can't do that myself as I don't have a Windows computer to try.

Can somebody make a well tested PR for this?
Is it even a reliable solution ?

[...]

PR #536, opened 2022-04-17
Issue #535 program not found on windows

Spacelord-XaN:
Added a cfg for windows for the following commands
* mkdir
* cp
* mv
* rm

Tested on Windows 11

The flags that are passed to the respective windows commands are open do discussion as I am not sure if they behave the same like in unix.


Canop:
Before I merge that, I need some windows users to check it.

Canop:
I got nobody to check that :\

I'll try to merge it anyway.

PR #582, opened 2022-08-06
RUSTSEC-2020-0159: upgrade chrono dependency from v0.4.19 to v0.4.20

creideiki:
Broot 1.14.2 depends on chrono = "0.4", which is currently fulfilled by 0.4.19.

chrono 0.4.19 is vulnerable to RUSTSEC-2020-0159: Potential segfault in localtime_r invocations

`cargo update --package chrono` bumps the dependency to chrono v0.4.20 and seems to work for me.


creideiki:
Upstream bumped dependency to 0.4.22.

PR #583, opened 2022-08-06
RUSTSEC-2020-0077: Switch from memmap to memmap2

creideiki:
memmap is unmaintained, according to RUSTSEC-2020-0077. memmap2 retains the same API.

I have done very limited testing, and only on Linux, but file content search and hex view works for me.


Canop:
I see no problem. Performances aren't easy enough to check on this (the changing cache of the SSD dwarfs all other considerations) but it looks ok.

Issue #504, opened 2022-02-15
token search broken under default directory (~)


nnsite:
Hi, yet another bug to report, yes sire !

when on my root repertory /home/drm/, and typing
`nt/imm`
broot crashes with this ![Image](https://imghost.org/images/2022/02/15/error_broot.md.jpg)
this bug is totally reproductive and consistent.
Again, sorry I can't seem to get the trace in any other way !
token searching works as long as I'm elsewhere than /home/drm/. Or, well, it ain't as reproductive and I don't remember or it didn't show up yet. I can provide bettre traces, if someone explains how :smile: 
The white boxes are my editing.

Canop:
Please indicate your OS and console and provide the [debug log](https://dystroy.org/broot/community/#log) of the crashing session.

evanescente-ondine:
Ah, I **knew** I missed something !!
`11:54:36.847 [INFO] cli_log::init: Starting broot v1.9.2 with log level DEBUG
11:54:36.847 [DEBUG] broot: env::args(): [
    "broot",
    "--outcmd",
    "/tmp/tmp.s9UN9FwUfb",
]
11:54:36.855 [DEBUG] broot::cli: Conf::from_default_location() took 7.073368ms
11:54:36.855 [DEBUG] broot::cli: config: Conf {
    files: [
        "/home/drm/.config/broot/conf.hjson",
    ],
    default_flags: Some(
        "h",
    ),
    date_time_format: None,
    verbs: [
        VerbConf {
            invocation: Some(
                "touch {new_file}",
            ),
            internal: None,
            external: None,
            execution: Some(
                String(
                    "touch {directory}/{new_file}",
                ),
            ),
            cmd: None,
            cmd_separator: None,
            key: None,
            keys: [],
            shortcut: None,
            leave_broot: Some(
                false,
            ),
            from_shell: None,
            apply_to: None,
            set_working_dir: None,
            working_dir: None,
            description: None,
            auto_exec: None,
        },
        VerbConf {
            invocation: Some(
                "edit",
            ),
            internal: None,
            external: Some(
                String(
                    "micro +{line} {file}",
                ),
            ),
            execution: None,
            cmd: None,
            cmd_separator: None,
            key: None,
            keys: [],
            shortcut: Some(
                "e",
            ),
            leave_broot: Some(
                false,
            ),
            from_shell: None,
            apply_to: None,
            set_working_dir: None,
            working_dir: None,
            description: None,
            auto_exec: None,
        },
[...]

PR #600, opened 2022-09-12
bugfix: fix `TokPattern::find_ranges` to solve #504

FedericoStra:
This addresses #504.


Canop:
Thanks a lot @FedericoStra

Issue #206, opened 2020-02-19
Stop printing line like "removed '/tmp/tmp.PnBgiBPflb'" after exiting broot


c02y:
**Is your feature request related to a problem? Please describe.**
Whenever I quit broot like `Ctrl-q` or `Ctrl-c` or simply using `broot --help` to print the usage message, there always be a line like "removed '/tmp/tmp.PnBgiBPflb'" printed, could you please remove this line?

Canop:
I've never seen such line.
Can you please show me the log of broot on a very short session ?
You can get it with
    BROOT_LOG=debug broot

c02y:
> BROOT_LOG=debug broot

Running `BROOT_LOG=debug broot` in bash, there is no "removed '/tmp/tmp.PnBgiBPflb'" line after quitting broot(using `Ctrl-q`).

dev.log file:
~~~
[chz@chz-pc tmp]$ less dev.log
13:17:53 [ INFO] Starting Broot v0.13.2 with log level DEBUG
13:17:53 [DEBUG] (1) broot::shell_install: Shell script already installed. Doing nothing.
13:17:53 [DEBUG] (1) broot::cli: Applying default flags "pdFI" from conf
13:17:53 [DEBUG] (1) broot::cli: modified tree options: TreeOptions { show_hidden: false, only_folders: false, show_sizes: false, show_dates: true, show_git_file_info: false, trim_root: true, show_permissions: true, respect_git_ignore: true, filter_by_git_status: false, pattern: None }
13:17:53 [DEBUG] (1) broot::app: we're on screen
13:17:53 [DEBUG] (1) broot::tree_build::builder: GitIgnorer::new took 388.059µs
13:17:53 [DEBUG] (1) broot::git_ignore: searching applicable gifs for "/home/chz/tmp"
13:17:53 [DEBUG] (1) broot::git_ignore:   looking in "/home/chz/tmp"
13:17:53 [DEBUG] (1) broot::git_ignore:   looking in "/home/chz"
13:17:53 [DEBUG] (1) broot::git_ignore:   looking in "/home"
13:17:53 [DEBUG] (1) broot::git_ignore:   looking in "/"
13:17:53 [DEBUG] (1) broot::tree_build::builder: Building - total=false pattern=None
13:17:53 [DEBUG] (1) broot::tree_build::builder: Trimming: we have 2 lines for a goal of 48
[...]

PR #629, opened 2022-11-04
Prevent `br` from using `rm` alias or function in Bash

jakubkaczor:
See https://github.com/Canop/broot/issues/206. I don't use nor know much about fish, so it would be nice if someone tested, if a similar change would work there as `command` is a shell built-in.

### DIFF
--- a/src/shell_install/bash.rs
+++ b/src/shell_install/bash.rs
@@ -43,11 +43,11 @@ function br {
     cmd_file=$(mktemp)
     if broot --outcmd "$cmd_file" "$@"; then
         cmd=$(<"$cmd_file")
-        rm -f "$cmd_file"
+        command rm -f "$cmd_file"
         eval "$cmd"
     else
         code=$?
-        rm -f "$cmd_file"
+        command rm -f "$cmd_file"
         return "$code"
     fi
 }


Canop:
Unsure.
This is interesting but what if people want to use their own modified `rm` ?

Canop:
Opinions welcome here.

jakubkaczor:
> Unsure. This is interesting but what if people want to use their own modified `rm` ?

Then they could modify the function by hand. Defaulting to using `command` avoids unexpected behavior.

kw-andy:
Hi @jakubkaczor , thanks for the `command`. It has been merged but thinking about it, what are the gains of using the `command`? It seems to prevent a type of errors that seems to me, but I maybe wrong, marginal.

Best

jakubkaczor:
> Hi @jakubkaczor , thanks for the `command`. It has been merged but thinking about it, what are the gains of using the `command`? It seems to prevent a type of errors that seems to me, but I maybe wrong, marginal.
> 
> Best

Hello @kw-andy, I'm not sure about which errors you talk about, if you can elaborate or give some example, I would be grateful. As I said, it avoids unexpected behavior. I will give one example.

Let us say one has `rm` aliased to `rm -v` to display information what is being done on interactive use. In this case, any use of `br` outputs information about the deletion of a temporary file on quit. This breaks using `br` in statements such as `$(br)`. Also, with regard to aesthetics, seeing the message about the deletion of some temporary file (implementation detail) on every quit may be unwanted, as mentioned in #206.

PR #647, opened 2023-01-01
Nushell br function

### LINKED ISSUE #375: Need a br shell function for nushell
An alias was previously proposed:

    alias br [] { broot | trim | cd $it }

but it doesn't seem to work now, either because of changes in nushell or because of changes in broot.

Help appreciated.

Canop:
Install the br function for nushell.

To have this function installed on a previously installed version of broot, run `broot --install`

Fix #375


mediumrarez:
Not working on Windows. `get_nushell_dir()` is returning the wrong path. It's supposed to be `%APPDATA%/nushell` on Windows. `~/.config/nushell/config.nu` is ignored, even if `%APPDATA%/nushell/config.nu` is deleted (tested).

Canop:
Is %APPDATA% the AppData/Roaming directory ? Is it [one of the directories BasesDir returns](https://docs.rs/directories/latest/directories/struct.BaseDirs.html) ?

FrancescElies:
Yes `%APPDATA%` = ~/AppData/Roaming`

Canop:
@FrancescElies Thanks. This should be an easy fix, then.

Canop:
@mediumrarez @FrancescElies Can you test again with the latest commit ?

FrancescElies:
Did a quick test with the latest commit.

image

I expected `RUST_BACKTRACE` to give me more context on which path it does not have privileges but it didn't.

Here my nushell dir structure
image

Canop:
> more context on which path it does not have privileges but it didn't

You're right. `shell_install/mod.rs` needs a better error handling. I'll work on that.

FrancescElies:
It was a permission problem in [this line](https://github.com/Canop/broot/blob/main/src/shell_install/mod.rs#L224) in `shell_install/mod.rs` you mentioned (normal users in windows are not allowed to create symlinks under windows), I am not sure what the best way would be to solve that problem, at the moment I just happen to be using windows but wouldn't adventure myself to say what to do about this, running `broot --install with admin rights did the following.
- Added a `source statement to my config.nu`
  image
- I opened broot via br on a new shell and seemed to work.

Canop:
Normal users are able to create files in this location but not symlinks ? Are symlinks something sensitive on Windows ?

FrancescElies:
¯\_(ツ)_/¯ I found [this](https://superuser.com/questions/10727/why-cant-normal-users-on-windows-create-symbolic-links) (my win knowledge is limited)

Canop:
@FrancescElies I've added context on errors, and some clarification for the user when that happens. Could you please remove the `config/broot/launcher `directory and launch the broot I just commited ?

FrancescElies:
The error message tells you what the files involved in the error 👍 great!
See below the test
image

Canop:
Damn, I didn't get the expected error...
Turns out "A required privilege is not held" isn't a "Permission denied".

I was expecting this kind of advice:
image

FrancescElies:
Sure, I'll give it a go tomorrow.

mediumrarez:
The latest commit is working just fine for me. It might be a good idea single-quote the file path in the sourcing line though.

As for the symlink permissions, I have some info to share. The symlink feature was introduced for NTFS in Windows Vista but restricted by admin privileges. That restriction was removed in one of the Windows 10 builds. If you're running a recent version of Windows it shouldn't be happening.

I don't know why @FrancescElies is having permission errors. Maybe the permissions for that folder itself are wrong?

FrancescElies:
> Could you try to fix this function: https://github.com/Canop/broot/pull/647/files#diff-6ff19a1ff611f3835b5a025663cf4dec901ed914626f78847414836439bd1a79R40 ?

This seems to catch the error https://github.com/FrancescElies/broot/commit/38b6d9b69aceacf9eb5c5cb6678554b149c0dcfd

Canop:
> This seems to catch the error [FrancescElies@38b6d9b](https://github.com/FrancescElies/broot/commit/38b6d9b69aceacf9eb5c5cb6678554b149c0dcfd)

Great. I don't see a way in GitHub UI to merge your change so I'll just copy it.

Canop:
Just waiting for final validation before merging in the PR.

PR #649, opened 2023-01-12
fix: nushell br alias fails on alt-enter if folder contains spaces

FrancescElies:
This is tiny fix for the case where a path contains white spaces in it.

Maybe @mediumrarez could double check what I am proposing is fine too?

### DIFF
--- a/src/shell_install/nushell.rs
+++ b/src/shell_install/nushell.rs
@@ -14,15 +14,10 @@
 
 use {
     super::{util, ShellInstall},
-    crate::{
-        conf,
-        errors::*,
-    },
+    crate::{conf, errors::*},
     directories::BaseDirs,
     std::path::PathBuf,
-    termimad::{
-        mad_print_inline,
-    },
+    termimad::mad_print_inline,
 };
 
 const NAME: &str = "nushell";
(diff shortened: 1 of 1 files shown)


mediumrarez:
It looks fine for the most part, but we should remove the `str replace -a '"' ''` line. Nushell is smart enough to deal with quotes by itself. It'll just cause path not found on linux when the path has `"` in in it.

FrancescElies:
@mediumrarez I agree with you, I couldn't test the `"` case (currently in a win. machine and it does not allow that character in the path) but I added a comment of how to test that code when doing changes to it.

Can you give it a try on your side?

Canop:
@FrancescElies You can add commits, no need to force push. I'll squash when merging if it's messy.

FrancescElies:
I am used to remove stuff in the history that adds nothing, but if you are checking out the branch you are right is more comfortable for you to just pull, I'll do like that from now on.

src/shell_install/nushell.rs
```diff
@@ -14,15 +14,10 @@
...
-        conf,
-        errors::*,
-    },
+    crate::{conf, errors::*},
     directories::BaseDirs,
     std::path::PathBuf,
-    termimad::{
-        mad_print_inline,
-    },
+    termimad::mad_print_inline,
```
FrancescElies, line 20: @Canop I had autoformat with cargo activated and I changed this lines by accident, this has nothing to do with this pr, shall I revert it or leave it as it is?

Canop: It's Ok, don't worry for such a small change, don't lose time reverting.

mediumrarez:
We could probably expand the regex to handle no-quotes and single quotes too. (Tested on Windows.)

```nushell
let target_dir = (open $cmd_file | parse -r `^cd\s+(?<quote>"|'|)(?<path>.+)\k<quote>[\s\r\n]*$` | get path | to text)
```

I'll have to test on linux tomorrow; it's too late for tonight.

src/shell_install/nushell.rs
Canop: BTW I'd like to use cargo fmt but things are slow to stabilize...
```
Warning: can't set `imports_layout = Vertical`, unstable features are only available in nightly channel.
Warning: can't set `version = Two`, unstable features are only available in nightly channel.
```

FrancescElies:
I will try it on Monday, but it looks good to me too, if you don’t mind shall I integrate it into this pr?

mediumrarez:
No, go ahead. Public domain 👌.

FrancescElies:
@mediumrarez if you are fine with the current code so am I :) and we could proceed for the merge

mediumrarez:
@FrancescElies Looks OK to me.

Canop:
OK, so I'll merge and try to have the new script replace the older one, right ?

FrancescElies:
I would say that's the right thing to do

Canop:
I'll add a system to have the script update proposed.

Canop:
https://github.com/Canop/broot/pull/655

PR #654, opened 2023-01-18
Dir preview

Canop:
- it's possible to call :open_preview when the selection is a directory
- when moving the selection in a browser view, the preview is moving accordingly even when it's a directory

Filtering, selections, scrolling, etc. are handled.
Like for other preview, the Enter key does nothing (for now).


n0bra1n3r:
Amazing! Is it possible to configure this using `default_flags` in the `toml` config file, so it's always on? I couldn't figure out how to use the `--cmd` switch from within the config file.

Canop:
@n0bra1n3r Why not change your launch shortcut to `br -c :open_preview` (you can define a function, depending on your shell, or even modify the /home/dys/.local/share/broot/launcher/bash/1 file) ?

PR #658, opened 2023-01-23
Add no-hide option to special paths

avlllo:
Hi there! In my projects I have several hidden folders that I need to see at all times. I noticed that there is a `hide` option for special paths, but no `no-hide` option. So this option has been added. I hope you'll find it useful.


Canop:
(sorry I'm not answered that before: I have some difficulties coping with all the work I have currently but I won't forget your PR)

Canop:
Very good PR. I merged it. My only concern now is that the limit that "special-paths" are only absolute paths will appear less relevant...

PR #666, opened 2023-02-07
Fix status not always displaying hint of input's verb

### LINKED ISSUE #665: Dangerously Incorrect rm verb statusline
If the file or directory selection is changed _after_ typing the `rm` verb, the statusline changes to give the incorrect instruction "Hit enter to open file" or "Hit enter to focus", when in-fact hitting enter will delete the file or directory.

This does not appear to be unique to the rm verb, it's just particularly dangerous with the rm verb. I'm not sure if the better solution is to fix the statusline, or change the behaviour... I'm leaning towards the latter, it's slightly less convenient but safer.

Canop:
Fix #665

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### next
+- fix status line not always displaying the hint of the input's verb - Fix #665
+
 ### v1.20.0 - 2023-02-03
 <a name="v1.20.0"></a>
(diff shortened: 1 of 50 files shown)


Canop:
I have a few checks to do before I merge that but anybody confirming it seems OK would be helpful.

ThomasBrierley:
Looks good. :+1:

Done a side by side comparison with my custom config. Only difference I could see is the info being correct after typing a verb and navigating.

PR #676, opened 2023-02-24
Fall back on /usr/bin/editor for $EDITOR

danielzgtg:
This makes broot use `/usr/bin/editor` when `$EDITOR` has not been set yet. This reduces friction and improves the new user experience.

`/usr/bin/editor` is the convention on Ubnutu. Many editors such as `/bin/ed`, `/usr/bin/vim.basic` and `/usr/bin/vim.tiny` among others use `update-alternatives` to offer themselves as providers for `/usr/bin/editor` via `/etc/alternatives/editor`. `$EDITOR` is not usually set on Ubuntu. In fact, the Ubuntu developers patched software specifically for this. [For example](https://packages.ubuntu.com/kinetic/bash), `debian/patches/bash-default-editor.diff` in `bash_5.2-1ubuntu2.debian.tar.xz` makes bash consider `/usr/bin/editor` when doing CTRL-x CTRL-e.

This PR implements this Ubuntu-like behavior. It will not change anything if `$EDITOR` is already set. If someone is relying on a literal command called `$EDITOR`, they can start with `export EDITOR='$EDITOR'`, using the absolute path, or something else to get that back.

# Before

```
Unable to launch $EDITOR: No such file or directory (os error 2)
```

# After

vim opens.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -204,6 +204,7 @@ dependencies = [
  "users",
  "usvg",
  "usvg-text-layout",
+ "which",
  "xterm-query 0.1.0",
 ]
 
@@ -2503,6 +2504,17 @@ name = "weezl"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9193164d4de03a926d909d3bc7c30543cecb35400c02114792c2cae20d5e2dbb"
+
+[[package]]
+name = "which"
+version = "4.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2441c784c52b289a054b7201fc93253e288f094e2f4be9058343127c4226a269"
+dependencies = [
+ "either",
+ "libc",
+ "once_cell",
+]
 
 [[package]]
 name = "winapi"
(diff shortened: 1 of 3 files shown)


Canop:
Thanks for the PR and research.

It doesn't feel right to do this inside broot but I see broot wouldn't be the first program to do it.

I'd welcome other inputs/opinions on this.

Canop:
This looks good.
I was also given an idea for an improvement: use the `which` crate to look for the path of `editor` rather than assuming it's `/usr/bin/editor`.

danielzgtg:
> use the which crate

Added

stevenxxiu:
Just to offer my opinion. Like @Canop I also disagree with this.

I don't think anyone expects there to be such a thing as `/usr/bin/editor` to even exist on a *Linux* distribution. Adding distribution specific code just seems strange to me. It's just forcing *Ubuntu* quirks onto everyone.

Canop:
@stevenxxiu With the added determination of the path of the `editor` command, and this being done only when `$EDITOR` is not set, I think it's in the category of ugly but useful little improvements. There's no real downside. I'm merging it with added logs.

PR #692, opened 2023-03-27
feat: add a `is_terminal` verb attribute, so externals that don't have terminal output, don't flicker

stevenxxiu:
I have a few custom external commands. One copies the current path with the home directory replaced with `~`. I use this one quite often. The constant flicker is rather annoying. Pretty much all of these commands flicker when run.

I discovered that the flicker is caused by `broot::launchable::Launchable::execute()` restoring the normal terminal for the command to run, and then reverting to *Broot*'s terminal. However none of the commands I use even have a terminal output.

So here I added a new `is_terminal` verb attribute. This defaults to `true` which is the previous behavior. But if set to `false`, it stops *Broot* restoring the normal terminal, hence removing the flicker

### DIFF
--- a/src/app/panel_state.rs
+++ b/src/app/panel_state.rs
@@ -973,7 +973,7 @@ pub trait PanelState {
     ) -> Status {
         if sel_info.count_paths() > 1 {
             if let VerbExecution::External(external) = &verb.execution {
-                if external.exec_mode != ExternalExecutionMode::StayInBroot {
+                if external.exec_mode != ExternalExecutionMode::StayInBrootTerminal {
                     return Status::new(
                         "only verbs returning to broot on end can be executed on a multi-selection".to_owned(),
                         true,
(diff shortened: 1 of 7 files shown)


Canop:
I'm missing something.
Can you please post one of your "custom external commands" as an example (in miaou if you prefer) ?

stevenxxiu:
Sure. Here's the worst case out of my custom external commands. I use it really often and it flickers every time.

```hjson
{
    key: ctrl-c
    external: ['sh', '-c', '~/.config/broot/copy_path.nu "$0"', '{file}']
    leave_broot: false
    is_terminal: false
}
```

Contents of `copy_path.nu`:

```
#!/usr/bin/nu
def _path_as_tilde [path: string] {
  $path | str replace $"^($env.HOME)\(?=/|$\)" ~
}

def main [path: string] {
  _path_as_tilde $path | xclip -in -selection clipboard
}
```

Canop:
@Stargateur I think I get it, the `is_terminal` flag is more like `switch_terminal`. I want to test a few things before I merge that, I just didn't have the time yet.

Canop:
BTW, your ctrl-c is like broot's `:copy_path` except that you want a tilde instead of the path to your home ?

PR #693, opened 2023-03-31
switch_terminal verb config parameter

Canop:
Broot runs in an "alternate terminal".
By default,  broot switches back to the normal terminal during the execution of a program it launches, then back to its alternate terminal on end of the external process.
In some cases, this adds an unnecessary flicker.
The `switch_terminal` verb parameter, whose default value is `true`, lets you disable this behaviour.

Same goal than https://github.com/Canop/broot/pull/692 but with some different data structures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### next
+- `switch_terminal` verb parameter - Thanks @stevenxxiu
+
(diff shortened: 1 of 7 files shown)


Stargateur:
`switch_terminal` is better name

PR #703, opened 2023-05-03
Correct some typos

goggle:
Corrects some typos in the repository.

### DIFF
--- a/src/tree/tree.rs
+++ b/src/tree/tree.rs
@@ -129,8 +129,8 @@ impl Tree {
             if self.lines[i].score > self.lines[best_index].score {
                 best_index = i;
             }
-            for d in 0..self.lines[i].left_branchs.len() {
-                self.lines[i].left_branchs[d] = false;
+            for d in 0..self.lines[i].left_branches.len() {
+                self.lines[i].left_branches[d] = false;
             }
         }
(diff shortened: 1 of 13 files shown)


src/tree/tree.rs
```diff
@@ -485,9 +485,9 @@ impl Tree {
                 // we'll try to keep the same path selected
                 let selected_path = self.selected_line().path.to_path_buf();
                 self.lines[1..].sort_by(|a, b| {
-                    let acount = a.sum.map_or(0, |s| s.to_count());
+                    let account = a.sum.map_or(0, |s| s.to_count());
```
Canop, line 488: It's not an "account", but the A count, just like there's a bcount.

src/verb/external_execution_mode.rs
```diff
@@ -3,7 +3,7 @@ pub enum ExternalExecutionMode {
     /// executed in the parent shell, on broot leaving, using the `br` function
     FromParentShell,
 
-    /// executed on broot leaving, not necessarly in the parent shell
+    /// executed on broot leaving, not necessary in the parent shell
```
Canop, line 6: The right one here would be "necessarily"

Canop (changes requested):
Thanks for that. A few changes are needed, though.

PR #706, opened 2023-05-17
left-elision of path when path/name doesn't fit

### LINKED ISSUE #700: shorten long directories to show filenames
Hi,

even on a maxed out terminal window im not able to see the full file names.

SCR-20230427-fv0

Sure this is a very insane folder structure, but sadly its a real world example and on smaler terminals this feature will make even more sense.
Here is how my shell is handling longer folder names:

SCR-20230427-fzy

For sure broot shall keep the matching parts, but this might be an idea to actually see something inside a horrible folder structure.

Just updated from 1.9 to the current version .. did i overlook an existing flag to already get this feature? If its already possible it should be a default, tho.

Canop:
Fix #700


Canop:
Checking it works even for characters taking several bytes and characters taking several screen cells:
image

PR #716, opened 2023-06-09
Add scoop to 3rd party repositories on the website docs

superusercode:
Scoop is a windows package manager, broot has been available through it for a while.


Canop:
Thanks.

As I don't know scoop at all, I'll let a few days for people to react here or in the chat before I merge the PR.

Canop:
I had a look myself. 

The [manifest](https://github.com/ScoopInstaller/Main/blob/master/bucket/broot.json) and the GH action bot update processes look good.

Issue #625, opened 2022-10-31
New install hooks (namely powershell and nushell)


melMass:
The title says it all!
I half solved the nushell one and did not look into the powershell one but it should be simple for it too.
I can look more into it soonish, creating this ticket as a reminder too.

Canop:
This kind of contribution is very welcome as I don't plan to invest time myself learning nushell or powershell in the near future.

I'll just remember that I want PRs to be focused (don't touch other files or lines), good quality, and that I always need to have testers to confirm the soundness of the approach according to the shell.

Canop:
> Do I need to know anything besides the single file per shell kind found in? src/shell_install

Well, you'll have to have it called, so you should probably read the whole `src/shell_install` module and the few lines calling it in `src/cli/mod.rs`

stinos:
See #159 :)

melMass:
@stinos Nice, thanks for the reminder, I had totally forget about that issue.

Nushell was still in the middle of changing stuffs that would ease this integration. 
These now shipped in 0.72+ so this can be tackled.

I quickly looked at your PS script and it looks nice, I will have time next week to address the `install` command of both if no one did by then.

stinos:
@melMass I'll keep an eye for your PR: I just made that script as a one-off and never really looked back but wouldn't mind reviewing it somewhat thoroughly. Like: that dotsource thing looks weird, needs a better implementation or at least a reference to why it's like that

Canop:
Both installers exist now: nushell since broot 1.19 (#375) and PowerShell since 1.24. Closing.

Issue #460, opened 2021-10-24
Launch broot as br in win 10


Trid-collab:
How do I launch broot as br in Windows 10

adriansosa:
I'll install Ubuntu 20.04 on wsl2 and report back if I have any issues.

thorstenkampe:
I just ran into this, too: when you run `broot` for the first time, it will ask to install a shell wrapper to launch `broot` as `br`. `broot` will install wrappers for `bash`, `zsh`, and `fish` (which are completely useless on Windows) but will not install a wrapper for PowerShell or Cmd. Despite of that it will claim success.

The fix is obvious: don't try to install the `br` wrappers on Windows (where they are useless) but notify the user that `cd` support is not implemented on Windows yet.

thorstenkampe:
Documenting the limited Windows support in the documentation would probably also lead to less frustration for Windows users.

whatisaphone:
`br` is very close to working on windows in git bash. If you run `br` and press alt+enter, broot writes this to the outcmd file:

```
cd C:\Users\John
```

The slashes are not quoted/escaped, so running that command results in this error:

```
sh: cd: C:UsersJohn: No error
```

If the path was escaped, it would work perfectly.

AeliusSaionji:
Here, I mocked this up. Windows users please test and let me know if there are issues.

Add it to your powershell profile (Run `echo $PROFILE` to see where the file should go if you don't have one yet, or `notepad $PROFILE` to jump right into it).

Note: on Windows Alt-Enter is swallowed by the terminal, used to make it fullscreen. You can unbind Alt-Enter with the new Windows Terminal, but not the built in ones. For those, you'll have to just use broot's `:cd` command directly.

```powershell
Function br {
  $args = $args -join ' '
  $cmd_file = New-TemporaryFile
  
  $process = Start-Process -FilePath 'broot.exe' `
                           -ArgumentList "--outcmd $($cmd_file.FullName) $args" `
                           -NoNewWindow -Wait -PassThru -WorkingDirectory $PWD

  If ($process.ExitCode -eq 0) {
    $cmd = Get-Content $cmd_file
    Remove-Item $cmd_file
    If ($cmd -ne $null) {
      Invoke-Expression -Command $cmd
    }
  } Else {
    Remove-Item $cmd_file
    Write-Warning "broot exited with error code $process.ExitCode"
  }
}

```

AeliusSaionji:
[...]

Issue #159, opened 2020-01-14
Add detection/launcher for Powershell


stinos:
In general it would be great if broot could detect which shell/environment it is running and install a suitable launcher; this issue in particular is for Powershell. Launchers have already been suggested in other issues, here's one take:

```
function br {
    $tempFile = New-TemporaryFile
    try {
        $broot = $env:BROOT
        if (-not $broot) {
             $broot = 'broot'
        }
        & $broot --outcmd $tempFile $args
        if ($LASTEXITCODE -ne 0) {
            Write-Error "$broot exited with code $LASTEXITCODE"
            return
        }
        $command = Get-Content $tempFile
        if ($command) {
            # broot returns extended-length paths but not all PowerShell/Windows
            # versions might handle this so strip the '\\?'
            Invoke-Expression $command.Replace("\\?\", "")
        }
    } finally {
        Remove-Item -force $tempFile
    }
}

$isDotSourced = $MyInvocation.InvocationName -eq '.' -or $MyInvocation.Line -eq ''
if (-not $isDotSourced) {
    br $args
}
```

- works as a script so if this gets saved as br.ps1 somewhere, users can choose how they want to use it (dot-source it, or add path to file to the PATH or ...)
- uses environment variable so users can set the path to the broot executable if broot isn't in the path for instance, or when developing and switching between versions etc. Defaults to broot instead of broot.exe so that it automatically works on pwsh on linux as well (tested on ubuntu 18.04)

Canop:
broot detects some shells, but not powershell, because I can't test that.

I'd need some win10 rust developer to try adapt what is done for fish: https://github.com/Canop/broot/blob/master/src/shell_install/fish.rs#L94

PDQDakota:
@Canop If it helps PowerShell is open source and cross platform. https://github.com/PowerShell/PowerShell#get-powershell

Canop:
@PDQDakota  Thanks, I didn't knew. It seems it's not the same version which is released with Windows, if I correctly understand the readme. And installing this on linux would probably not lead to the typical installation I want to detect.
Moreover, I don't have the time to learn this shell, so I'd like a Rust Windows developer to look at this.

GrosSacASac:
Last time I tried, I compiled broot with the rust compiler then used the full path to the compiled broot inside powershell

pshirshov:
PowerShell Core works fine on MacOS and Linux. It would be really great if you add support for it.

Canop:
[...]

PR #720, opened 2023-07-10
Powershell br function install

felixkroemer:
A PowerShell br function install module based on [AeliusSaionji's br function](https://github.com/Canop/broot/issues/460#issuecomment-1303005689) as suggested in #625 and #159.  It follows the same conventions as the other install modules. The function is added to the $HOME\Documents\WindowsPowerShell\Profile.ps1 profile.

I tested this on two Windows 10 machines but not on Windows 11.

### DIFF
--- a/src/shell_install/mod.rs
+++ b/src/shell_install/mod.rs
@@ -3,6 +3,7 @@ mod fish;
 mod nushell;
 mod state;
 mod util;
+mod powershell;
 
 use {
     crate::{
@@ -87,6 +88,7 @@ impl ShellInstall {
             "bash" | "zsh" => println!("{}", bash::get_script()),
             "fish" => println!("{}", fish::get_script()),
             "nushell" => println!("{}", nushell::get_script()),
+            "powershell" => println!("{}", powershell::get_script()),
             _ => {
                 return Err(ProgramError::UnknowShell {
                     shell: shell.to_string(),
@@ -135,6 +137,7 @@ impl ShellInstall {
         bash::install(self)?;
         fish::install(self)?;
         nushell::install(self)?;
+        powershell::install(self)?;
         self.should_quit = true;
         if self.done {
             self.skin.print_text(MD_INSTALL_DONE);
--- a/src/shell_install/powershell.rs
+++ b/src/shell_install/powershell.rs
@@ -1,0 +1,116 @@
+use std::fs;
+
+use directories::UserDirs;
+
+use {
+    super::{util, ShellInstall},
+    crate::{conf, errors::*},
+    std::path::PathBuf,
+    termimad::mad_print_inline,
+};
+
+const NAME: &str = "powershell";
+const VERSION: &str = "1";
+
+const PS_FUNC: &str = r#"
+# https://github.com/Canop/broot/issues/460#issuecomment-1303005689
+Function br {
+  $args = $args -join ' '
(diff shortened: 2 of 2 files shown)


Canop:
Testers welcome.

Canop:
Just tested it on a Windows computer and your change seems to perfectly work.

There are a few hard points though, not sure where to put the necessary explanations, and not sure if they affect all users:
* I had to run the shell as administrator for the installation to run
* I had to enable scripts with `set-executionpolicy unrestricted`

Advice welcome on the best way to make this easier or clearer for normal users.

FrancescElies:
Not a windows pro here (just having to use it from time to time).

Also Not sure if the following would be right for broot but fnm solves it by like [this](https://github.com/Schniz/fnm#using-scoop-windows) and then adds a section about [powershell](https://github.com/Schniz/fnm#powershell).

One needs to call something like `fnm env --use-on-cd | Out-String | Invoke-Expression` inside the script that fnm spits out to the stdout one could have something like `Start-Process powershell -Verb runAs -ArgumentList "echo my-elevated-command; sleep 10"` to have a command that is executed with elevated privileges.

AeliusSaionji:
Thanks for adding this!

I do have a couple notes:

---

Bug: If `Documents\WindowsPowerShell` already exists, the command to create it fails.
Improvement: Also make a symlink for pwsh 😄 `Documents\PowerShell\profile.ps1`, possibly asking the user if they actually want to if the folder doesn't already exist.

---

> I had to enable scripts with set-executionpolicy unrestricted

Restricted execution policy will be the default on all end user systems. Users will need to change this to even execute the powershell profile.ps1
We can advise inexperienced users to use RemoteSigned rather than unrestricted, and to do it for their user account only with:
```powershell
set-executionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
```

As noted here, locally written scripts need not be signed, and the current method of generating the script via rust would probably not have the "remote" markings on it.
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-7.3#manage-signed-and-unsigned-scripts

---

> I had to run the shell as administrator for the installation to run

Administrator access is indeed required to make symlinks, but that requirement is dropped if you have Developer Mode enabled.

---

> One needs to call something like fnm env --use-on-cd | Out-String | Invoke-Expression

To be clear, this is a suggestion that broot pivot from the current scripts (which write --outcmd to a file) and instead do something like `alias br=broot --outcmd | <shell specific eval>` ? The one problem I see with that is, what if there is no stdout to be evaluated?

felixkroemer:
> Bug: If Documents\WindowsPowerShell already exists, the command to create it fails.

Thanks. I missed that one. Fixed it [here](https://github.com/Canop/broot/pull/723).

> Oddity: br sometimes asks me if I want to install br 🤔

That is weird. I believe it should put the marker file into the config directory after the installation. Does it ask only once after a successful install?

AeliusSaionji:
Yeah, only once, but I seemed to have reset the state a few times by poking at the --install

AeliusSaionji:
Hm, I can reproduce the problem on any system and OS

```bash
broot --install
br
```

Running the above, br will always ask to install, every time.

Canop:
@AeliusSaionji Can you move this problem to a new issue ? This would be easier to follow than in a closed PR's discussion

PR #755, opened 2023-10-09
Add extra shortcuts for builtins

9999years:
- input_go_word_left: alt-b
- input_go_word_right: alt-f
- line_down: ctrl-j
- line_up: ctrl-k
- panel_left_no_open: ctrl-h
- panel_right_no_open: ctrl-h

These don't conflict with any existing builtin shortcuts and users can override them if they want.


Canop:
Those look a little too personal, apart maybe the `line_down` and `line_up`. Where do the `alt-f` for `input_go_word_right` come from ?
Unless there's some support from other users I'm reluctant to add those: if they're just personal, they can be just in your config.

Other users are invited to give their opinion.

9999years:
`alt-b` and `alt-f` for word-back and word-forward come from Emacs, although they're supported in many terminals as well.

cckuan2718:
> `alt-b` and `alt-f` for word-back and word-forward come from Emacs, although they're supported in many terminals as well.

Is it possible to have readline(3) like binding when in command mode? Helix editor did it.

Canop:
The right solution to me seems to be the addition of a commented-out block of settings in the default configuration, similar to what we already have for vim-like shortcuts: https://github.com/Canop/broot/blob/main/resources/default-conf/verbs.hjson#L113

PR #758, opened 2023-10-11
Print the root path relative to the initial root

9999years:


Canop:
Can you please explain what you want to achieve and why ?

9999years:
Yes, I'll mark it ready for review once I have an explanation :)

9999years:
Relative paths are nice and I want to see them instead of absolute paths most of the time.

I'm also developing [a `broot` integration plugin for Neovim](https://github.com/9999years/broot.nvim/pull/14) and I'm using a snapshot testing framework. However, I can't get the tests to pass locally and in CI because I'm running them in different directories. If I can just view relative paths this isn't a problem.

I could also [patch the snapshot testing framework to be able to ignore specific regions of the screenshots, as I've proposed here,](https://github.com/echasnovski/mini.nvim/issues/512) but I wouldn't like to have to manually draw out those regions.

PS Thank you very much for Broot, it's very nice!

Issue #760, opened 2023-10-12
`cd` to Git repo root


9999years:
Hello! I'd like to be able to quickly change my directory to the Git repository root, e.g. with a `gr` verb.

I could define a verb with an `internal = ":focus {git-root}"`, but there's no way for me to get the `{git-root}` value. It's easy enough to call `git rev-parse --show-toplevel`, but I don't believe the `internal` verb attribute supports shell execution.

Should we implement this specific feature (e.g. a `{git-root}` substitution variable) or a more general one (e.g. `$(...)` shell interpolation in `internal` or `external` verb definition attributes)?

Canop:
This is an interesting question but I don't think there's a need to change broot for this.

I defined this function in my `.bashrc`:

```
function go_to_git_root() {
    cd $1
    cd $(git rev-parse --show-toplevel)
}
```

Then I added this verb definition in my `verbs.hjson`:

```
    {
        invocation: gr
        external: ["go_to_git_root", "{directory}"]
        from_shell: true
    }
```

It seems to do what you want.

9999years:
Oh, but I don't want to leave broot, I want to `:focus` the repo root for further searching.

Canop:
Oh, sorry, I had misunderstood you.

partiallyordered:
Woo! I can't believe I came here to make this very issue (after having it in my to-do list for quite some time) and here it is right at the top of the list.

While I'm here, thank you @Canop, the more I use Broot, the better it gets! (And of course, the better you make it, the better it gets!).

I recognise this is a verbose +1, hope it's not too annoying but you are very welcome to delete.

9999years:
I made an attempt at this in #761.

9999years:
Closed with #761 merged.

PR #761, opened 2023-10-13
Add `{git-root}` substitution

9999years:
Initial attempt at #760. This works well in my testing.

### DIFF
--- a/src/verb/internal_focus.rs
+++ b/src/verb/internal_focus.rs
@@ -194,10 +194,6 @@ pub fn on_internal(
                 input_arg,
                 app_state,
             );
-            let tree_options = TreeOptions {
-                pattern: InputPattern::none(),
-                ..tree_options
-            };
             on_path(path, screen, tree_options, bang, con)
         }
         _ => {
--- a/src/verb/internal_focus.rs
+++ b/src/verb/internal_focus.rs
@@ -1,5 +1,7 @@
 //! utility functions to help handle the `:focus` internal
 
+use crate::pattern::InputPattern;
+
 use {
     super::*,
     crate::{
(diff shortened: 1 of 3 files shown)


src/verb/internal_focus.rs
```diff
@@ -122,7 +124,13 @@ fn path_from_input(
             // The given path may be relative hence the need for the
             // state's selection
             // (we assume a check before ensured it doesn't need an input)
-            path::path_from(base_path, PathAnchor::Unspecified, verb_arg)
+            let path_builder = ExecutionStringBuilder::with_invocation(
```
9999years, line 127: Here we use the `ExecutionStringBuilder` so that we can use the `{...}` substitutions in arguments to the `focus` verb.

src/verb/internal_focus.rs
```diff
@@ -186,6 +194,10 @@ pub fn on_internal(
                 input_arg,
                 app_state,
             );
+            let tree_options = TreeOptions {
+                pattern: InputPattern::none(),
+                ..tree_options
+            };
```
9999years, line 200: When `focus` is triggered by a verb, clear the input pattern. The idea is you're e.g. looking for a Git repo, so you type in some substring to get it selected, and then you trigger a `focus {git-repo}` verb which focuses you on the repo root and clears your filter pattern so you're ready to look around. But I think this is a change from the current behavior, so I'm not sure if you'll want this.

9999years: It might be nicer to have a `clear_filter` option for verb definitions to control this behavior and then have some extra state in `CmdResult` for it.

Canop: I know there's a UX problem there and I have in my TODO list to clarify this and implement something like this (I hadn't thought about a visible `clear_filter` option, this is actually a good idea).

Note: please don't attempt to code this clear_filter thing, I prefer for now to handle those core changes myself.

9999years: OK, reverted this hunk.

Canop:
@9999years Sorry for the lack of review: I'm very busy ATM, I have a huge backlog (including several things on broot) but I didn't forget, I'll get to your PR and issue.

PR #769, opened 2023-10-19
Extend 'apply_to' with 'text_file' and 'binary_file'

### LINKED ISSUE #764: Automatically open text files with `$EDITOR`
Terminal file managers like `nnn`, `lf`, `yazi` can automatically detect text files and open them in your `$EDITOR`, instead of `xdg-open`. `broot` requires you to specify an explicit key for this, I think it'd be more convenient if we can use a single key to open anything.

Canop:
This makes it possible, for example, to have text files open in the same terminal on 'enter':

```Hjson
{
    invocation: edit
    key: enter
    shortcut: e
    execution: "/usr/bin/nvim +{line} {file}"
    apply_to: text_file
    leave_broot: false
}
```

Fix #764

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### next
+- the `apply_to` verb filter accepts new values: `text_file` and `binary_file`. Broot users editing files in their terminal (vi, emacs, etc.) should configure broot to open their text editor on `enter`: see https://dystroy.org/broot/tricks/#change-standard-file-opening
 - small breaking change: `:stage_all_files` now stages also symlinks - Fix #606
 - new `{git-root}` verb argument - Fix 760 - Thanks @9999years
 
--- a/resources/default-conf/verbs.hjson
+++ b/resources/default-conf/verbs.hjson
@@ -6,23 +6,10 @@
 
 verbs: [
 
-    # Example 1: launching `tail -n` on the selected file (leaving broot)
-    # {
-    #     name: tail_lines
-    #     invocation: tl {lines_count}
-    #     execution: "tail -f -n {lines_count} {file}"
-    # }
-
-    # Example 2: creating a new file without leaving broot
-    # {
-    #     name: touch
-    #     invocation: touch {new_file}
-    #     execution: "touch {directory}/{new_file}"
-    #     leave_broot: false
-    # }
-
-    # A standard recommended command for editing files, that you
-    # can customize.
+    # You should customize this standard opening of text files.
+    # If you edit text files in your terminal (vi, emacs, helix, eg.), then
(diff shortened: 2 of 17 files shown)


musjj:
I'm trying to build this, but for some reason I keep getting this error:

```rust
       > error[E0425]: cannot find function `is_file_known_binary` in module `magic_numbers`
       >   --> src/content_type/mod.rs:27:20
       >    |
       > 27 |     magic_numbers::is_file_known_binary(path)
       >    |                    ^^^^^^^^^^^^^^^^^^^^ not found in `magic_numbers`
       >
```

Is it a typo of this function?:

https://github.com/Canop/broot/blob/aead8106a202d900439be867c11a0c29a55565aa/src/content_search/magic_numbers.rs#L76

Canop:
But the function is in the PR files: https://github.com/Canop/broot/pull/769/files#diff-875aa6ab1b4041dd7920d66de1999be6a1af4e3c5c00d467211ec70015ec543bR120

Canop:
@musjj if it looks easier for solving this problem, please come to the chat https://miaou.dystroy.org/3490

musjj:
Weird, I was applying this PR as a patch on top of the main branch, so I decided to just build the feature branch directly and now it works. I guess something went wrong with my nix derivation.

I tested the new `apply_to: text_file` and it seems to work perfectly!

Canop:
I'll keep the branch unmerged for tests for a week, I think.

Among the downsides, it's harder to open SVG files as images.

musjj:
I'll use this as a daily driver and report any issues I may encounter. 

So far I don't really mind svg being opened as text files, since I can just type ` os` if I really want to open an image viewer/editor. Most of the time I'm interacting with svg files in the terminal is when I'm trying to edit them as a text file, but I guess that varies between users.

But maybe if you want it to be more flexible, how about something like a new `disallowed_extensions` option (the opposite of `extensions`)?

Canop:
> But maybe if you want it to be more flexible, how about something like a new disallowed_extensions option (the opposite of extensions)?

I must be cautious or I'll end up with a scripting language just to decide what verbs apply...

jinliu:
Works for me. And perhaps this should be in the default config file.

Canop:
> And perhaps this should be in the default config file.

I'm still unsure about that: it means the enter key won't work for people without a $EDITOR. And it won't suit people whose editor isn't a terminal one.

FrancescElies:
A bit late to help but this works fine for me under windows+nushell, default binding `ctrl-e` opens nvim just fine.

Canop:
> A bit late to help 

A late report is much better and more appreciated than no report. Thanks.

Issue #588, opened 2022-08-24
Alt-Enter does not work on Windows


philippseith:
alt-enter is the Windows Console / Terminal shortcut to maximize/restore the console/terminal window. Therefore, it is not possible to execute :open_leave or :cd with this shortcut. 
Changing the shortcut works for :open_leave only. When trying to set
```hjson
{
  key: shift-enter
  execution: ":cd"
}
```
you get `Bad configuration: not a known internal: cd` 

this stand in the way of a  `br` function in  powershell, where you can change dir with a shortcut:
```pwsh
function br() {
	$tmp = [System.IO.Path]::GetTempFileName()
	c:\OwnApp\Tools\broot.exe --outcmd "$tmp" $args
	$cd = Get-Content "$tmp"
	Remove-Item -Force "$tmp"
	$dir = $cd.substring(3)
	if (Test-Path -PathType Container "$dir") {
		if ("$dir" -ne "$pwd") {
			Push-Location "$dir"
		}
	}
}
```
(selecting the path and then `:cd` is working, but a shortcut would be nice)

This problem occurs with broot running locally on Windows, but also with broot running on a remote host, when connected over ssh in the Windows Console / Terminal

philippseith:
In Windows Terminal, it is possible to remove the Alt-Enter shortcut in Settings.

AeliusSaionji:
new to broot.

On a related note, the help page within broot references `open_leave` and the shortcut `ol`, as working on directories. That's not the case even on linux.

Because `open_leave` so thoroughly caught my eye and had a visible sister command `open_stay`, I managed to overlook the existence of `:cd` for several hours and therefore thought the dropping back into tho shell could be invoked /only/ by alt+enter.

Seems like the command should be `open_leave` and `cd` should be the shortcut, rather than having two commands which do the same thing according to `:help` (but they don't) 🤪

Canop:
> On a related note, the help page within broot references open_leave and the shortcut ol, as working on directories. That's not the case even on linux.

Hum... I'll investigate this

Canop:
Ok, found what happens.

![image](https://user-images.githubusercontent.com/617006/200126978-3506b31f-8816-42b7-b16d-5bc72a1b1350.png)

The "ol" shortcut is mapped to :open_leave but only for files (i.e. not directories):

        // those two operations are mapped on ALT-ENTER, one
        // for directories and the other one for the other files
        internal(open_leave) // calls the system open
            .with_stype(SelectionType::File)
[...]

Issue #99, opened 2020-01-09
Trouble installing on MSYS2/MINGW64


runvnc:
Hello so I have a cargo install on my Windows computer and it seems like it works under MSYS2 Mingw64 but after running my Mingw64 terminal under Administrator and then trying broot and telling it to complete the br part it says this:

```
*runvnc@DESKTOP-O0LSE3G MINGW64 ~
# broot

Broot should be launched using a shell function (see https://github.com/Canop/broot for explanations).
The function is either missing, old or badly installed.
Can I install it now? [Y n]
Y
Removing C:\Users\runvnc\AppData\Roaming\dystroy\broot\data\launcher\bash\1.
Writing br shell function in C:\Users\runvnc\AppData\Roaming\dystroy\broot\data\launcher\bash\1.
Creating link from C:\Users\runvnc\AppData\Roaming\dystroy\broot\config\launcher\bash\br to C:\Users\runvnc\AppData\Roaming\dystroy\broot\data\launcher\bash\1.

I found neither .bashrc nor .zshrc.
If you're using bash or zsh, then installation isn't complete:
the br function initialization script won't be sourced unless you source it yourself.
```
So it looks like I need to manually figure out how to source the "br function initialization" and then add that to my .bashrc myself.  So I think that means source the 'br' script but I am not sure.

runvnc:
Well I sourced the br script, it is giving the same message when I run it as br.  I think that maybe for my configuration or just in general this isn't set up to work with Mingw.

Canop:
@runvnc I suspect this is the same problem than https://github.com/Canop/broot/issues/98 at the root

FrancescElies:
Not sure if this is an option for you but broot works well with nushell (which comes with many gnu coreutils itegrated) in windows, aparently there are more people happy with that setup see #460.

FrancescElies:
Not sure if this is an option for you but broot works well under windows with nushell (which comes with many gnu coreutils [integrated](https://www.nushell.sh/blog/2023-09-05-why-uu.html)), it looks like there are more people happy with that setup see #460.

AeliusSaionji:
I have written a PKGBUILD for compiling and distributing under msys2, but I haven't submitted it upstream at the time of writing.

- Must manually update pkgver to target the desired release
- This does not install the br function, and really shouldn't anyway. That would ideally be fixed within broot. However you can manually `broot --print-shell-function bash >> ~/.bash_profile`.

```bash
# aelius
_pkgname=broot
[...]

Issue #98, opened 2020-01-09
Fails to start in Msysgit bash on Windows


sidfarkus:
Trying to use this on Windows 10 Pro Version 10.0.18362 Build 18362

Installing with Rust:
` cargo install broot`

And adding broot to bashrc:
`source /c/Users/myuser/AppData/Roaming/dystroy/broot/config/launcher/bash/br`

Running `br` gives me the following output:
```
Getting mode failed
Termimad Error : "Crossterm error"
```

When I run `broot` in powershell on my system it functions but is slow to redraw.

Canop:
I asked Timon, the leader of the crossterm project.
It's a known problem: 

> We can't call into WinApi from cygwin terminals.  I'm not sure how to solve that problem. Since we have to use it for input reading for example.

sidfarkus:
Fooey; it does sound like it's difficult to tie those things together.  I think this setup just might not be supported for a while then.

Canop:
@sidfarkus There's still hope. I reopen the issue as the problem is clear and is absolutely not solved.

carloscm:
This combination *almost* works:
- Windows 10 cmd.exe
- Use cygwin64
- Use ConEmu x64
- Use the broot.exe builds from github
1) Start a cmd.exe prompt inside ConEmu
2) Inside the cmd.exe prompt, invoke bash.exe from cygwin
3) Inside that bash.exe you can then launch broot.exe, with the following issues:
- It will output Win32 UNC paths instead of the cygwin.dll mapped paths (this is expected, since broot.exe is not compiled with the cygwin toolchain)
- It doesn't actually change the current folder for the bash session
- It refreshes the screen slowly
- The first line of output is always cut out

The first and second issue are what will probably make this a futile goal, since there is no native cygwin build of the Rust toolchain as far as I am aware. So even if the screen I/O issues are solved, broot.exe will use Win32 paths/APIs and cygwin shells will use the cygwin.dll equivalents, which are not compatible and/or invisible to each other.

Canop:
Don't worry for the "spam". It's better to record it. Even if I don't really have time now to study this, somebody else may, now or later.

gadros:
1. can't use it in bash - also get the rust crossterm error. frustrating.
2. when ran in the new windows terminal or powershell can't 'cd' into any folder because alt+enter toggles the window between maximized or normal size (not easy work around). renders it unusable
[...]

Issue #78, opened 2020-01-07
Paths on Windows are invalid


71:
**Describe the bug**
On Windows, the extended path with UNC prefix are used to canonicalize paths (e.g in the root directory at the top, or in output of commands). This causes issues

**To Reproduce**
Using the following Powershell Core function, try to `:cd` into a directory.

```pwsh
function br
{
    $TempFile = New-TemporaryFile
    broot.exe --outcmd $TempFile $Args

    if (!$?)
    {
        Remove-Item -Force $TempFile
        return $LASTEXITCODE
    }

    $CommandToExecute = Get-Content $TempFile
    Remove-Item -Force $TempFile
    Invoke-Expression $CommandToExecute
}
```

An error will be shown, indicating that the path `\\?\A:\Foo` does not exist (even though `A:\Foo` does exist).

**Expected behavior**
Normal (usable) links with no UNC prefix should be used, since most commands do not support those prefixes.

**Configuration**
- OS: Windows
- Version: 10.0.18362

**Additional context**
As first reported [here](https://github.com/Canop/broot/issues/50#issuecomment-571059434), the problem is that Broot [calls](https://github.com/Canop/broot/blob/d327ae6cb0fd00b66f4ca33664b904684269e9e0/src/cli.rs#L151) [`std::fs::canonicalize`](https://doc.rust-lang.org/std/fs/fn.canonicalize.html), which returns a path like `\\?\C:\Users\71`, instead of `C:\Users\71`. Therefore using `:cd` inside Broot will write a command `cd '\\?\A:\foo'`, which will confuse PowerShell (and most commands in general).

A workaround is to execute `$Command.Replace('\\?\', '')` instead of `$Command`.

In the end, the script I ended up using is:

```pwsh
function br
{
    $TempFile = New-TemporaryFile
    broot.exe --outcmd $TempFile $Args

    if (!$?)
    {
        Remove-Item -Force $TempFile
        return $LASTEXITCODE
    }

    $CommandToExecute = Get-Content $TempFile
    Remove-Item -Force $TempFile
    Invoke-Expression $CommandToExecute.Replace('\\?\', '')
}
```

**However**, it looks like this **does not work** in all cases. [This crate](https://gitlab.com/kornelski/dunce#dunce-de-unc) attempts to make the transformation in a smart way, but I wonder what should happen when it judges that the UNC prefix cannot be stripped.

Canop:
[...]

PR #771, opened 2023-10-26
MVC to fix path issues on Windows

3tilley:
I'm a Windows user who mainly uses bash. I can see there are some rough edges on Windows, and it's noted elsewhere that the maintainers aren't Windows users themselves (which is understandable). I'm hoping to try and work through some of the rough edges to do with path handling, especially with bash.

I can see #588, #460, #78, #98, #99 which should mostly be fixed by something similar to the below. Have you got any others?

I'll have to do some work to formally detect cygwin / bash on windows, but I'll hardcode it locally for myself and run it for a while to see how it goes.

I wonder if the more elegant solution is to use / for paths everywhere instead of quoting, but that will bring up its own issues. I'll see how it works with the quoting for a while and then think about it a bit more.

### DIFF
--- a/src/verb/execution_builder.rs
+++ b/src/verb/execution_builder.rs
@@ -409,5 +409,10 @@ mod execution_builder_test {
 }
 
 fn path_to_string<P: AsRef<Path>>(path: P) -> String {
-    path.as_ref().to_string_lossy().to_string()
+    let is_bash_on_windows = true;
+    if is_bash_on_windows {
+        format!("\"{}\"", path.as_ref().to_string_lossy().to_string())
+    } else {
+        path.as_ref().to_string_lossy().to_string()
+    }
 }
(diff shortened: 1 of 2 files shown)


src/verb/execution_builder.rs
```diff
@@ -409,5 +409,10 @@ mod execution_builder_test {
 }
 
 fn path_to_string<P: AsRef<Path>>(path: P) -> String {
-    path.as_ref().to_string_lossy().to_string()
+    let is_bash_on_windows = true;
```
Canop, line 412: Did you forget something here ?

3tilley: Yes, deliberately so! I have hardcoded it locally to allow me to test `broot` in Windows bash, and when I've done some testing I'll implement the best method.

This is more of a CTA for any similar bugs that I can fix in one shot.

PR #786, opened 2023-11-08
fix: ✨ nushell deprecation

melMass:
def-env is now def --env

image

### DIFF
--- a/src/shell_install/nushell.rs
+++ b/src/shell_install/nushell.rs
@@ -33,7 +33,7 @@ const NU_FUNC: &str = r#"
 #   > br -hi -c "vacheblan.svg;:open_preview" ..
 #
 # See https://dystroy.org/broot/install-br/
-def-env br [
+def --env br [
     --cmd(-c): string               # Semicolon separated commands to execute
     --color: string = "auto"        # Whether to have styles and colors (auto is default and usually OK) [possible values: auto, yes, no]
     --conf: string                  # Semicolon separated paths to specific config files"),


Canop:
Ok, found the reference: https://www.nushell.sh/blog/2023-10-17-nushell_0_86.html#unified-command-definitions

melMass:
> Did you check the script works with `def --env` ? What about users of older nushell ?

Yep this works from older versions, I just asked on nushell's discord since when, but they usually deprecate first and remove later.

def --env is supported in the previous release (0.85) for instance (the warning from my screen will only happen in 0.86+)

melMass:
Actually no, it seems I was wrong... I probably installed a main version of nushell before the release so the version stated 0.85 but this actually landed in 0.86 so to answer your question it's not backward compatible.

First appeared here: https://github.com/nushell/nushell/pull/10566
Deprecation added here: https://github.com/nushell/nushell/pull/10715 & https://github.com/nushell/nushell/pull/10716

Canop:
They look quite aggressive with their deprecation... How are we supposed to write a highly compatible script without dumping error messages ?

melMass:
> How are we supposed to write a highly compatible script without dumping error messages ?

I asked exactly that after discovering it was "abruptly" deprecated. 
The main reason seems to be that they only have error mechanism to properly show where issue occurs in source and not more subtle warning that would fit better here. 

In the end it is probably safer to not merge this and people with newer version will have to deal with the warning/error for the time being, I'll keep pushing for a more subtle error.

The true deprecation is planned for 0.88

Canop:
Note: this is **not** closed.

It's pending until either there's a solution to the retro-compatibility problem or it looks like the right moment to switch to the new version (and I'll need the input for nushell users to trigger this).

FrancescElies:
> Did you check the script works with `def --env` ? What about users of older nushell ?

I recently switched to nushell v0.87.1, and `def --env` works fine there.

> In the end it is probably safer to not merge this and people with newer version will have to deal with the warning/error for the time being, I'll keep pushing for a more subtle error.

👍 thanks

happysalada:
The latest version of nushell that deprecates the command is now released, it would be amazing if this can get merged.

Canop:
Okey

Canop:
This is released in 1.30.1

happysalada:
Thank you !

PR #791, opened 2023-11-11
Better default flags

Canop:
Before this commit, only the one letter flags interpreted
as tree options were supported.

Now
- the -c (or --cmd) argument is also supported
- long form arguments are supported

Examples:

    default_flags: "-s -c thing -gh"
    default_flags: "gh" // old syntax
    default_flags: --show-git-info --only-folders


website/docs/conf_file.md
```diff
@@ -87,10 +87,10 @@ For example, if you want to see hidden files (the ones whose name starts with a
 If you almost always want those flags, you may define them as default in the configuration file file, with the `default_flags` setting.
 
 ```Hjson
-default_flags: gh
+default_flags: -gh
```
cohml, line 90: I feel this could be clearer.

It kind of *implies* that you can specify a full string of flags inside the `"..."` exactly as you would on the CLI (e.g., `broot --flag1 --flag2=val -3` --> `default_flags: "--flag1 --flag2=val -3"`). But it would be really nice if that were stated explicitly, either in the prose description above, or via a more complex example.

website/docs/conf_file.md
```diff
@@ -87,10 +87,10 @@ For example, if you want to see hidden files (the ones whose name starts with a
...
 ```
 ```TOML
-default_flags = "gh"
+default_flags = "-gh"
```
cohml, line 93: Ditto, obv.

cohml (changes requested):
Just requesting minor improvement to docs. Explicit > implied, IMHO.

cohml:
By the way, my immediate needs are met, but I did have a thought: You updated [the docs]() with this line:

> The `default_flags` entry lets you specify them in configuration, with the same syntax.

However, the same information is not contained in the actual config file. From `conf.hjson` (v1.28.1):

> # Default flags
> # You can set up flags you want broot to start with by
> # default, for example `default_flags="ihp"` if you usually want
> # to see hidden and gitignored files and the permissions (then
> # if you don't want the hidden files you can launch `br -H`)
> # A popular flag is the `g` one which displays git related info.

JFYI in case you wanted to update that as well.

Issue #788, opened 2023-11-10
`br` command fails on Windows


computerscience-person:
After installing `broot` using scoop and executing `broot --install` to be able to use the `br` command, I'm met with an error.
```
> br
File not found: "C:\\Users\\Username\\AppData\\Local\\Temp\\tmpC170.tmp"
```
I think it's because my username has a space in it, as it duplicated the second word in my username. I looked at the modified `$PROFILE`, and I'm not sure what's wrong with the `br` function, though I'm not versed in `.ps1` syntax.

Canop:
Is that the exact error message ? With "Username" ?

computerscience-person:
No, just changed my username since it contained my actual name.

Canop:
The problem is probably in the way the shell script gives the path to broot but I don't know powershell well enough and I don't have any Windows box available so I need somebody else to try fix the script.

lost22git:
https://github.com/Canop/broot/blob/4856d360c83ca0ac6a341afa577ec3e875de113c/src/shell_install/powershell.rs#L29C1-L29C1

Modify `$($cmd_file.FullName)` to 
``` 
`"$($cmd_file.FullName)`"
```
maybe work

Elkiwa:
I tried to reproduce this by creating a test user but can't even complete the `br` function install prerequisite. I wonder what should I do besides having a space in username? 

Steps:
1. [Create a local user account](https://support.microsoft.com/en-us/windows/create-a-local-user-or-administrator-account-in-windows-20de74e0-ac7f-3502-a866-32915af2a34d)
2. Deliberately input the user name "test use"
3. Install broot and try to install `br` function with `sudo broot --install`
```PowerShell
PS C:\Users\test use> sudo broot --install
You requested a clean (re)install.
Removing C:\Users\test use\AppData\Roaming\dystroy\broot\data\launcher\bash\1.
Writing br shell function in C:\Users\test use\AppData\Roaming\dystroy\broot\data\launcher\bash\1.
Removing C:\Users\test use\AppData\Roaming\dystroy\broot\config\launcher\bash\br.
Creating link from C:\Users\test use\AppData\Roaming\dystroy\broot\config\launcher\bash\br to C:\Users\test use\AppData\Roaming\dystroy\broot\data\launcher\bash\1.

I found no sourcing file for the bash/zsh family.
If you're using bash or zsh, then installation isn't complete:
the br function initialization script won't be sourced unless you source it yourself.
PS C:\Users\test use>
```
[...]

PR #793, opened 2023-11-12
fix github issue #788

lost22git:


Canop:
@lost22git Did you test it ?

Daverd-GM:
I tried to compile it, it compiled. but it's still not generating the correct path in the .PROFILE file
Captura de pantalla 2024-02-05 133812
Now it supports me to put the correct path and it loads correctly 
Captura de pantalla 2024-02-05 142512
so now the app supports the right path but you still need to change it by yourself after running
'''ps
broot --install 
'''
in admin mode

Issue #670, opened 2023-02-15
Toggle to show only items in current directory


jtourlamain:
Is there a toggle to show only items in the current directory (folders and files)? Showing content of all sub-directories can produce to much info. Having a toggle to only show the current items in the directory and search for them would be great.

usama8800:
All sorts except sort by type do what you want https://dystroy.org/broot/tree_view/#sort:~:text=There%20are%204%20kinds%20of%20sort%3A

jtourlamain:
@usama8800 the problem with the sort is that is slow (it needs to calculate counts or dates or ... and on windows takes a second or 2, but that is to slow when navigating around). If there is a sort_by_name (which seems to be the default) but without displaying the sub directories, it would be fast and without the sub directories.

yusuphgammut:
Just want to upvote the `:sort_by_name` suggestion. It'll be an explicit way of showing only the files and folders on the current directory. If this sounds redundant, maybe adding a `:toogle_tree` command to just hiding or showing the tree can be an alternative solution.

eldad:
Same here. It bothers me so much I implemented it in my private fork:
https://github.com/Canop/broot/compare/main...eldad:broot:toggle_tree

It's quite a simple toggle, and I would be happy to contribute as a PR if you're open for this feature.

Canop:
@eldad This seems quite good and I think a PR would be welcome (please make the formatting consistent [in panel_state](https://github.com/Canop/broot/compare/main...eldad:broot:toggle_tree#diff-6b167eac73a6c185407141207531632ceb5b6be7d8dc9743cfa78b3e72d441b0R320).

I'm not quite convinced by the `toggle_tree` name but I don't have any better one at the moment.

eldad:
@Canop Thanks for the quick feedback! I'll brush up the changes, fix the formatting as you pointed out, and open a PR shortly.

Canop:
Done since 1.29 with the :toggle_tree internal and the --tree/--no-tree launch flags

PR #800, opened 2023-11-17
Add `toggle_tree` flag to show only items in the current dir

eldad:
This PR aims to address issue https://github.com/Canop/broot/issues/670

Thanks for creating and maintaining this tool!

### DIFF
--- a/src/app/panel_state.rs
+++ b/src/app/panel_state.rs
@@ -315,6 +315,21 @@ pub trait PanelState {
 					con,
 				)
             }
+            Internal::toggle_tree => {
+                self.with_new_options(
+                    screen,
+                    &|o| {
+                        o.toggle_tree ^= true;
+                        if o.toggle_tree {
+                            "*allow displaying tree structure*"
+                        } else {
+                            "*prevent tree structure*"
+                        }
+                    },
+                    bang,
+                    con,
+                )
+            }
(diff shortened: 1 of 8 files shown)


eldad:
This commit unintentionally changed the default behavior to not display the tree, I am pushing a commit to change the default behavior as it is currently.

eldad:
The last commit is now the desired behavior, default to show the tree.

Canop:
Having "toggle" in a launch argument isn't semantically right: most other arguments are toggles anyway. It also doesn't say whether it toggles towards enabling a tree view or disabling it.

IMO the right semantic is what we have for other toggles: a couple of launch arguments `--tree` and `--no-tree`. This would also make it possible to have `--no-tree` in `default_flags` in the config and to override it with `--tree` at launch.

(another option would be to just don't have a launch parameter as we can do `--cmd :toggle_tree`)

If you prefer, I can do the change myself starting from your last commit.

eldad:
> Having "toggle" in a launch argument isn't semantically right: most other arguments are toggles anyway. It also doesn't say whether it toggles towards enabling a tree view or disabling it.

Yes, I agree. That is not intuitive.

> If you prefer, I can do the change myself starting from your last commit.

I added now the change as you suggested - please feel free to edit it further as you see fit.

PR #803, opened 2023-11-29
:trash internal

### LINKED ISSUE #799: Optionally use system trash implementation when using `rm`
When using the `rm` verb, optionally move the deleted items to the system trash instead of unlinking them.

Nushell supports this with `rm`. It appears to use this implementation: https://docs.rs/trash/latest/trash/

Canop:
Add the `:trash` internal, which instead of deleting a file moves it to the system trash.

You may remap `rm` to it with 

    {
        invocation: "rm"
        internal: "trash {file}"
        leave_broot: false
    }

Fix #799

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -265,6 +265,7 @@ dependencies = [
  "terminal-clipboard",
  "terminal-light",
  "toml",
+ "trash",
  "umask",
  "unicode-width",
  "uzers",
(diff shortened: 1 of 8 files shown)


FrancescElies:
Late report is better than no report, windows nushell 0.87.1, works fine, internal trash command moves file to trash.

Canop:
> Late report is better than no report

100%. Thanks

Issue #804, opened 2023-11-30
For a symlink to another partition, opening it in a new panel shows a wrong fs size


jinliu:
STEPS TO REPRODUCE:
1. Create a symlink to a dir in another partition.
2. `broot --show-root-fs`
3. Press `right` on that link.
4. The fs size on the header bar updates correctly.
5. Press `left` to go back.
6. Press `Ctrl+right` to open the symlink in a new panel.
7. The fs size in the new panel is still that of the left panel.

jinliu:
Perhaps one solution is to make `ctrl-right` behave like `right`: the current directory is set to the target of the symlink, not the symlink itself.

Canop:
I reproduce the problem. I'll have a look.

jinliu:
Yes, it works. Thanks!

Canop:
BTW I'll push another change, I think: removing the filter when doing :panel_right on a directory

alexn-s:
i think this fix or at least version 1.30.0 > breaks something on windows @Canop .
the right/ second panel is broken and commands like cpp and mvp wont work.

the path is not correct

### version 1.30.0, 1.30.1
![image](https://github.com/Canop/broot/assets/60710873/605d0aee-a5c0-4949-b6ad-b3d976eced47)

### version 1.29.0
![image](https://github.com/Canop/broot/assets/60710873/c00a853c-688c-401a-a230-c11a1acc957d)

### sys info
OS: Windows 11 Home x86_64
Terminal: Windows Terminal
shell: nushell

let me know if i can provide more information

PR #810, opened 2023-12-22
don't canonicalize paths on windows on new panels

### LINKED ISSUE #809: actions on new panel broken on windows since 1.30
This was reported by @alexs-n here: https://github.com/Canop/broot/issues/804#issuecomment-1866958386

> i think this fix or at least version 1.30.0 > breaks something on windows @Canop .
the right/ second panel is broken and commands like cpp and mvp wont work.

> the path is not correct

image

Canop:
Fix #809

It's thus possible that issue #804 exists on Windows, I can't test that.

### DIFF
--- a/src/verb/internal_focus.rs
+++ b/src/verb/internal_focus.rs
@@ -13,7 +13,6 @@ use {
         tree::TreeOptions,
     },
     std::{
-        fs,
         path::{Path, PathBuf},
     },
 };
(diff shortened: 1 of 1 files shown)


Elkiwa:
This fixed the issue on my Windows!
PS. Windows Version 10.0.22631 Build 22631

Canop:
@Elkiwa Thanks for confirmation. I'll merge the fix, then

PR #814, opened 2023-12-29
Sync broot working dir with current panel's root

### LINKED ISSUE #813: Keep working directory of broot process in sync with directory of its currently focused panel
Hello. I'd like to suggest to keep broot's main process working directory in sync with the directory it is currently displaying just like the famous **nnn** file browser does (where you can literally just `ls -la /proc/<pid of specific nnn process>/cwd` to get the currently opened directory in that specific nnn instance).

This would have immediate benefits at least in:
- Zellij (terminal multiplexer) where a new pane gets opened in the working directory of the process running in the currently focused pane.
- Alacritty (terminal) where _SpawnNewInstance_ command opens a new window in the working directory of the current TTY controlling process of the current window.

(Caveat: Zellij seems to be using CWD of the oldest still running process in the currently focused pane so **nnn** (or broot in the future) must be running in that pane directly without underlying shell like bash or fish (which is quite normal and common anyway in Zellij). Alacritty is more inteligent and uses CWD of the current TTY controlling process.)

Canop:
Fix #813

I'm not really sure there are many use cases for this, and I might gate it behind a configuration flag if I suspect there might be unwanted side effects.

### DIFF
--- a/resources/default-conf/conf.hjson
+++ b/resources/default-conf/conf.hjson
@@ -168,6 +168,15 @@ content_search_max_file_size: 10MB
 # max_panels_count: 2
 
 ###############################################################
+# Update work dir
+#
+# By default, broot process' work dir is kept in sync with the
+# current's panel root. If you want to keep it unchanged,
+# uncomment this setting
+#
+# update_work_dir: false
+
+###############################################################
 # Imports
 #
 # While it's possible to have all configuration in one file,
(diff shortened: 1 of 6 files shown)


WizardUli:
I've tested this PR as requested in https://github.com/Canop/broot/issues/813#issuecomment-1872480463 and it seems to be working as expected.
- When two tree panels are displayed the CWD is correctly synced to currently focused one.
- When one of two panels is the staging area the CWD is correctly synced to the tree panel (focused or not).
- Displaying help screen does not change the CWD as expected.

Can't think of any more possible edge-cases since I'm very new to broot.
Nice job! :)

PR #822, opened 2024-01-16
Update nushell.rs to fix error on exit

LudoPinelli:
_Nushell_ deprecated the use of automatic spreading lists in favor of the `...` syntax, which throws an error each time _broot_ launched via `br` is closed.

This PR just update the syntax accordingly.

### DIFF
--- a/src/shell_install/nushell.rs
+++ b/src/shell_install/nushell.rs
@@ -113,7 +113,7 @@ def --env br [
     let cmd_file = ([ $nu.temp-path, $"broot-(random chars).tmp" ] | path join)
     touch $cmd_file
     if ($file == null) {
-        ^broot --outcmd $cmd_file $args
+        ^broot --outcmd $cmd_file ...$args
     } else {
         ^broot --outcmd $cmd_file $args $file
     }
--- a/src/shell_install/nushell.rs
+++ b/src/shell_install/nushell.rs
@@ -115,7 +115,7 @@ def --env br [
     if ($file == null) {
         ^broot --outcmd $cmd_file ...$args
     } else {
-        ^broot --outcmd $cmd_file $args $file
+        ^broot --outcmd $cmd_file ...$args $file
     }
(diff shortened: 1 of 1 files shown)


src/shell_install/nushell.rs
```diff
@@ -113,7 +113,7 @@ def --env br [
     let cmd_file = ([ $nu.temp-path, $"broot-(random chars).tmp" ] | path join)
     touch $cmd_file
     if ($file == null) {
-        ^broot --outcmd $cmd_file $args
+        ^broot --outcmd $cmd_file ...$args
     } else {
         ^broot --outcmd $cmd_file $args $file
```
FrancescElies, line 118: shouldn't we do the same in this line?

Canop:
Guys, I don't have time to test nushell ATM. When you both think the PR is finished and OK, you tell me and I merge.

src/shell_install/nushell.rs
LudoPinelli: Absolutely, I didn't try `br` with a command, and it logically also throws the error:
```
Error:   × Automatically spreading lists is deprecated
    ╭─[/home/ludo/.config/local/share/broot/launcher/nushell/5:92:1]
 92 │     } else {
 93 │         ^broot --outcmd $cmd_file $args $file
    ·                                   ──┬──
    ·                                     ╰── Spreading lists automatically when calling external commands is deprecated and will be removed in 0.91.
 94 │     }
    ╰────
  help: Use the spread operator (put a '...' before the argument)
```

LudoPinelli: I took the time to test it in every condition I can think of: without argument and with every argument `br` can have, and didn't have anymore error.
Do you think on a specific use case we should test, or should we give the go?

FrancescElies: I think we are good to go

FrancescElies:
@Canop we can merge

PR #826, opened 2024-01-18
support of multi-keys combinations, termimad 0.28

Canop:
With this PR comes the ability on terminals implementing [Kitty keyboard extensions](https://sw.kovidgoyal.net/kitty/keyboard-protocol/) to bind verbs to key combinations which are normally not available on simple ANSI terminals, for example `space-n`, `ctrl-enter`, `shift-space`, `ctrl-alt-a-b-c`,  etc.

As a lot of things changed in the engine (i.e. in the crokey and termimad libraries), this PR comes also with risks of regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
+### next
+- new syntax for special paths - Fix #687, #669
+
 ### v1.32.0 - 2024-01-02
 <a name="v1.32.0"></a>
-- whith "modal" enabled, `initial_mode` setting lets you choose whether to start in `input` mode or `command` mode (default) - Fix #708
+- with "modal" enabled, `initial_mode` setting lets you choose whether to start in `input` mode or `command` mode (default) - Fix #708
 
 ### v1.31.0 - 2023-12-30
 <a name="v1.31.0"></a>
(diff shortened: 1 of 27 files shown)


jinliu:
When enabled, this introduces conceivable delay with function keys (up, down, pgup, pgdn), on Kitty master (https://github.com/kovidgoyal/kitty#commit=4a64f812ad046b4aea023b85a0ce258da2c5572e).

Canop:
Do you see a bigger delay than for other keys ?

I wonder if I shouldn't change the approach: instead of waiting for the key release, I could could react on key press unless there's a modifier (ctrl, alt, shift, maybe space) key pressed. This would prevent multi-key combinations without modifier key but would be faster for standard keys (including function keys).

jinliu:
I feel all keys have delays. Even when entering verbs, where I think very little processing is happening. My estimate is ~100-200ms.

With the option disabled, I don't feel any delay at all.

Canop:
100-200ms ? Damn... That's absolutely not the key release problem then. I'll have to try with the last kitty to see if I reproduce.

Canop:
It should be better now (still no explanation for such a delay, but most combinations are now built on key press).

jinliu:
It works. Now the latency is barely noticable.

I can still feel a tiny bit of it. But it may well be mental. I'd add some timer code between keypress and screen actually updating to verify.

Canop:
Current blocking problem: I can't manage to get the caret key on wezterm (it works fine on kitty).

Canop:
There are other problems in wezterm, eg the ':' key.
See https://github.com/wez/wezterm/issues/4785
So I prefer to make the support of this protocol opt-in rather than opt-out.

Issue #825, opened 2024-01-17
Add a --verb-output launch argument


Canop:
Similar to the existing `--outcmd`, but the file whose path is given as argument would contain what the verb decides to  write.

The verb could for example be `:write {file}:{line}` or `:write {path}`.

Each time there's a write after the first one, a `\n` would be inserted.
There could be several lines, for example when a verb doesn't quit broot, or when the verb is called in the staging area.

Such a file would mainly be the basis for a rewritten br shell function, which would solve some escaping problems.

Related: https://github.com/Canop/broot/issues/824

AndydeCleyre:
Thanks! To be clear, in your example, `:write` is a new internal that appends to the file specified by `--verb-output`?

Like a manual logger?

Canop:
> Thanks! To be clear, in your example, :write is a new internal that appends to the file specified by --verb-output?

Yes.

AndydeCleyre:
FWIW, I happen to be using a verb with that particular name:

```yaml
  {
    invocation: write {subpath}
    external: $EDITOR {directory}/{subpath}
    leave_broot: false
  }
```

And I'm inclined to think of the upcoming internal as "append line"

Canop:
For a verb similar to yours, which is dedicated to creating new files, I use "create" with shortcut "cr".

`append_line` wouldn't really convey the intent, which is most often to write only one path and exit.

AndydeCleyre:
No! Thanks for the work and the ping!

I'll test today or tonight.

PR #834, opened 2024-02-11
:write_output & :clear_output, --verb_output

Canop:
Fix: #825

Add a `--verb-output` launch argument which takes a path to a file (which will be created if necessary)

Add a  `:write_output` internal which allows adding a line to that file. No escaping is done (contrary to what happens with `--outcmd`).

Add a `:clear_output` internal which clears the file.

Here are 2 examples of verbs:

```
    {
        invocation: "wc {cmd}"
        execution: ":write_output wc:{cmd} {file-stem}.bro"
    }

    {
        key: alt-w
        cmd: ":clear_output;:write_output {directory};:quit"
    }
```

The first one is called with an input like `:wc hop` which appends to the output a line like `wc:hop main.bro`. 

The second one makes the content of the output file the directory closest to the selection then quits. It could for example be used for a new version of the `br` shell function.

Note: Semantics isn't pretty. If you have a better idea than "output", please tell me.

### DIFF
--- a/src/app/panel_state.rs
+++ b/src/app/panel_state.rs
@@ -755,7 +755,7 @@ pub trait PanelState {
         seq_ex: &SequenceExecution,
         invocation: Option<&VerbInvocation>,
         app_state: &mut AppState,
-        _cc: &CmdContext,
+        cc: &CmdContext,
     ) -> Result<CmdResult, ProgramError> {
         let sel_info = self.sel_info(app_state);
         if matches!(sel_info, SelInfo::More(_)) {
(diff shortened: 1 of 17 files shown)


AndydeCleyre:
Thanks so much!

In my first tests, this seems almost working right, but for some reason the following verb does not quit broot. If I call the verb then manually quit it does seem to work as intended though.

```yaml
  {
    key: alt-enter
    invocation: cd
    internal: ":write_output {directory}"
    leave_broot: true
  }
```

Canop:
Today, internals don't even check the `leave_broot` setting of the verb. For most of them, quitting would be totally unexpected (think for example about "line_down", or "open_preview").

That's why my example in the PR description had an explicit `:quit`.

Maybe *some* internals could check it ?

AndydeCleyre:
I've been using `:print_path` with `leave_broot: true`, but I guess leaving broot is part of that internal's behavior anyway? Haven't tried it with `false`...

OK yeah I'll make it a cmd with quit, thanks!

Canop:
This works for me

AndydeCleyre:
Oh, I tried to use it on a folder named

`with " a quote and spaces`

Canop:
Ok... I suspected there was a bug in sequence interpolation... back to draft mode!

Canop:
@AndydeCleyre back to testable

AndydeCleyre:
This seems to be working great with my weird folder names, thanks!

Canop:
Thanks @AndydeCleyre for the review

Issue #789, opened 2023-11-10
Image preview in kitty is slow


dev-ardi:
(On kitty) Image preview takes much longer than one would expect. 

If you hold down the down key in an image directory broot lags incredibly hard.

leira:
I'm on Wezterm. The image preview with `br` is significantly slower than [yazi](https://github.com/sxyazi/yazi) and [ranger](https://github.com/ranger/ranger). To load a 6000*4000 picture, yazi was almost instant. ranger loaded within half seconds. It took around 5 seconds with br. How can I provide more information?

Canop:
@dev-ardi @leira can you try the version of [this PR](https://github.com/Canop/broot/pull/835) ?

PR #835, opened 2024-02-11
use temp_file transmission medium for kitty img

Canop:
Use one of those settings:

```
kitty_graphics_transmission: chunks
kitty_graphics_transmission: temp_file
```

The faster `temp_file` mode is the current default. Downside is it won't work for a remote broot execution.

May solve #789


Canop:
A temporary buid of broot has been made for testers: https://dystroy.org/broot/download/broot-faster-kitty-renderer

Issue #838, opened 2024-02-15
Get relative to git root file path to be used in verbs


VasilisManol:
I have this verb now:
{
        invocation: copy_filename
        key: ctrl-f
        execution: ["wl-copy", "{file}:{line}"]
        leave_broot: false
}

I would like to have something like {file-git-relative} to get the git-root relative path.

PR #841, opened 2024-02-19
Add file-git-relative verb argument fix#838

VasilisManol:
implements #838

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ### next
 - don't fail launch in case of bad verb configuration, more helpful error message in such case
 - faster kitty image rendering by default - Fix #789
-
+- `{file-git-relative}` verb argument
+ 
 ### v1.33.1 - 2024-02-03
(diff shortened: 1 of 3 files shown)


Canop:
When the current root is a git repository and you select it, the current value is an empty string. Wouldn't it be more useful to have `.` instead ?

Similar question when the current root is the parent of a git repository and you select it: the current value is the path to that directory. I think a value of  `.` would be better too.

VasilisManol:
For my use case I would like it to work as `{file}` does, unless into a git repo where I want the relative path. I define it in a shortcut to be able to get references to a file:line and then send it to colleagues. Therefore, it is convenient to be relative to git root. For all the other cases I want it to be like `{file}`. What would be the use case, where I would like to get a `.`? I can't think of any.
So my preferences for your questions would be the same, get the full path to the directory. ( same as `{file}` ).

Canop:
This will be included in the next release, hopefully this week-end

PR #842, opened 2024-02-22
write installed flag file on --install

### LINKED ISSUE #837: broot --install followed by br asks to install again
My `.zshrc` (on Mac) starts with:

```sh
[[ ! -e $XDG_CONFIG_HOME/broot/launcher/bash/br ]] && broot --install
source "$XDG_CONFIG_HOME/broot/launcher/bash/br"
```

But the first time I run `br` it asks:

```
Broot should be launched using a shell function.
This function most notably makes it possible to cd from inside broot
(see https://dystroy.org/broot/install-br/ for explanations).

Can I install it now? [Y/n]
```

The cause appears to be the implementation checking for `$XDG_CONFIG_HOME/broot/launcher/installed-vX` but `broot --install` doesn't include it.

Canop:
Fix #837

### DIFF
--- a/src/shell_install/mod.rs
+++ b/src/shell_install/mod.rs
@@ -127,11 +127,11 @@ impl ShellInstall {
                     }
                 }
             }
-            // even if the installation isn't really complete (for example
-            // when no bash file was found), we don't want to ask the user
-            // again, we'll assume it's done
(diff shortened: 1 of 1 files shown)


texastoland:
Great catch! I wondered about that line when I rereported it 🫣

Confirmed working on my build :shipit:

PR #865, opened 2024-04-06
add support for setuid, setgid and sticky bits

### LINKED ISSUE #863: broot fail to recognize `suid` and `sgid` bits when showing permission
This is a comparison between `broot` and `lsd`, both showing the same file `mount.cifs`, which has permission `rwsr-sr-x`:

comparison between `broot` and `lsd`

The two `s` in permission means `setuid` and `setgid` respectively, but broot does not recognize those bit and only shows `x` instead.

Jisu-Woniu:
closes #863

### DIFF
--- a/src/display/permissions.rs
+++ b/src/display/permissions.rs
@@ -7,10 +7,7 @@ use {
         skin::StyleMap,
         tree::{Tree, TreeLine},
     },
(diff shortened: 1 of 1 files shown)


Canop:
This looks good, thanks

PR #866, opened 2024-04-06
optionally show lines surrounding a match in preview

### LINKED ISSUE #756: Context lines for content matches
When searching file content with `c/...`, I can open the preview pane and then cycle through matches, but I only get the exact line that matched. I'd really like to have the matching line centered in a regular preview and to be able to cycle through the matches that way. Alternatively, having *n* lines of context before and after the match (like `rg`) would be nice.

Canop:
New settings:

```
###############################################################
# lines around matching line in filtered preview
#
# When searching the content of a file, you can have either
# only the matching lines displayed, or some of the surrounding
# ones too.
#
lines_before_match_in_preview: 1
lines_after_match_in_preview: 1
```

Fix #756

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### next
+- optionally display lines surrounding a matching one in display, with `lines_before_match_in_preview` and  `lines_after_match_in_preview` - Fix #756
+
 ### v1.36.1 - 2024-03-11
 <a name="v1.36.1"></a>
 - fix ANSI code leaking to the input on start on Mac - Fix #854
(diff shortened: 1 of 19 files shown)


stevenxxiu:
Thanks for this. Could there be a horizontal line break of sorts, between search results that aren't next to each other?

Canop:
Can you check if everything looks right ?

note: there's no separator when lines_before and lines_after are 0, to keep it dense.

stevenxxiu:
Thanks! Looks good to me.

Canop:
I need to support the tab to go to next match

Canop:
Done. Now you can go to next match with `tab` and the previous one with `shift-tab` (with default bindings).

This is needed since the down and arrow keys don't jump to the next/previous matching line.

Canop:
Question is... Should I set `lines_before_match_in_preview` and `lines_after_match_in_preview` to `0` or `1` by default ?

stevenxxiu:
Up to you I suppose.

*Ripgrep* itself doesn't enable this by default.

Setting this to 0 will keep the previous behavior, and 1 will help people notice this feature.

Personally I set this to 2.

PR #874, opened 2024-04-30
make it possible to use flags like '-sdp' as verb

Canop:
For example, to show dates and permissions:
image

Opinions on this feature welcome.


Canop:
With this PR comes a new verb invocation syntax, allowing to have verb with name such "-".

Here's the doc:

```
    /// Examples:
    ///  "mv"       -> name: "mv"
    ///  "!mv"      -> name: "mv", bang
    ///  "mv a b"   -> name: "mv", args: "a b"
    ///  "mv!a b"   -> name: "mv", args: "a b", bang
    ///  "a-b  c"   -> name: "a-b", args: "c", bang
    ///  "-sp"      -> name: "-", args: "sp"
    ///  "-a b"     -> name: "-", args: "a b"
    ///  "-a b"     -> name: "-", args: "a b"
    ///  "--a"      -> name: "--", args: "a"
    ///
    /// Notes:
    /// 1. A name is either "special" (only made of non alpha characters)
    ///    or normal (starting with an alpha character). Special names don't
    ///    need a space afterwards, as the first alpha character will start
    ///    the args.
    /// 2. The space or colon after the name is optional if there's a bang
    ///    after the name: the bang is the separator.
    /// 3. Duplicate separators before args are ignored (they're usually typos)
    /// 4. An opening parenthesis starts args
```

FrancescElies:
This is useful, thanks!

PR #877, opened 2024-05-07
fix builds on android

### LINKED ISSUE #861: Does not compile on Termux (Android) any more
Since #803 introduced a dependency on the `trash` crate, broot does not compile any more, presumably because Android does not have a trash.

```
error[E0433]: failed to resolve: use of undeclared crate or module `platform`
  --> /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-6f17d22bba15001f/trash-3.3.1/src/lib.rs:67:35
   |
67 |         Self { platform_specific: platform::PlatformTrashContext::new() }
   |                                   ^^^^^^^^ use of undeclared crate or module `platform`

error[E0433]: failed to resolve: use of undeclared crate or module `platform`
  --> /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-6f17d22bba15001f/trash-3.3.1/src/lib.rs:63:24
   |
63 |     platform_specific: platform::PlatformTrashContext,
   |                        ^^^^^^^^ use of undeclared crate or module `platform`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `trash` (lib) due to 2 previous errors
```

Ideally trash support would be behind a feature so that there is a way to opt out of this feature on platforms which do not have a trash. The feature could be on by default since most major platforms do, but at least it would be easy to make builds on platforms that do not.

dead10ck:
Since #803 introduced a dependency on the trash crate, broot does not compile any more on Android. This change makes the trash crate only pulled in on platforms which support it, and otherwise the trash command displays an error.

Fixes #861

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,10 @@ exclude = ["website", "broot*.zip"]
 
 [features]
 default = []
+
 clipboard = ["terminal-clipboard"]
 kitty-csi-check = ["xterm-query"]
+trash = ["dep:trash"]
 
 [dependencies]
 ahash = { version = "0.8.3", features = ["serde"] }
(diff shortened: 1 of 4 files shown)


Canop:
From the test output, it looks like there are places missing some `#[cfg(feature=trash)]` guard.

dead10ck:
Looks like attributes from the build script can't be used for dependency resolution, so it wasn't pulling in the dependency

Canop:
Thanks, merged.

Issue #827, opened 2024-01-26
nushell: `cd` not working


Schweber:
I am using broot 1.32.0 with nushell 0.89.0 and applied the argument expansion to the `br` shell function from https://github.com/Canop/broot/issues/821

However, `:open_leave` doesn't work for me in nushell. I've set `{ key: "enter", execution: ":open_leave" }` but if i press enter on a file/directory in broot, i'm back in the terminal without the folder having changed. Running `pwd` confirms this, so it's not a problem of the prompt not being updated.

Using `{ invocation: "cd", key: "enter", execution: "$SHELL", set_working_dir: true }` "works" but this launches an additional instance of `nu` with the given folder so it's not a good workaround.

What could be the reason for this?

Canop:
Just to be sure: you launch `br,` not `broot` ?

Canop:
`:open_leave` isn't supposed (or able) to change the current directory of the opening shell.

Try with this:

    {
        key: enter
        internal: open_leave
        apply_to: file
    }
    {
        key: enter
        external: "cd {directory}"
        from_shell: true
        apply_to: directory
    }

Canop:
> Is there a way to get a verbose output of broot to see why it fails? I found no such cli option...

Yes, there's a log: https://dystroy.org/broot/community/#log

Canop:
You don't have other verbs than the one I proposed mapped to the enter key ?

I'll have a deeper look and try on nushell but it won't be before a few days.

FrancescElies:
@Schweber I think this is spected because of nushell [scoping rules](https://www.nushell.sh/book/environment.html#defining-environment-from-custom-commands), the environment changes inside your def function stay there and do not have side effects outside the funciton, try with `def --env a [] { br -gis }`.

`def --env` is also what we used to define the br wrapper for nushell, see [here](https://github.com/Canop/broot/blob/main/src/shell_install/nushell.rs#L36)

Does this help?

PR #880, opened 2024-05-25
website: amend `install-br.md` with a note for nushell

Schweber:
https://github.com/Canop/broot/issues/827

### DIFF
--- a/website/docs/install-br.md
+++ b/website/docs/install-br.md
@@ -14,7 +14,8 @@ When it's done, you can do `br` to launch broot, and typing <kbd>alt</kbd><kbd>e
 Supported shells today are bash, zsh, fish, nushell, and powershell.
 
 !!! Note
-	**Windows users:** broot may need additional rights at first use in order to write its configuration file. You may also have to allow script execution (`set-executionpolicy unrestricted`)
+	**Windows users:** broot may need additional rights at first use in order to write its configuration file. You may also have to allow script execution (`set-executionpolicy unrestricted`)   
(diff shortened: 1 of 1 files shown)


Canop:
@FrancescElies @LudoPinelli @texastoland Your opinion on this would be welcome.

texastoland:
I'm traveling this week and haven't read the other thread completely. I agree the difference between `alias` and `def` here is surprising and worth documenting. Worth mentioning in the Nushell Discord server too in case it can be made more predictable. Tag me if you do 👍🏼

texastoland:
I mean you you identified an issue with Nushell not broot. Even though it works by design it was confusing regardless. If lazygit added a `lg` wrapper command should they document this behavior too?

The [XY problem](https://xy-problem.com/) is it wasn't intuitive that `br` works by changing the environment thus necessitating wrapping with `def --env`. Possible solutions include adding a warning to the compiler, making it more obvious in the LSP (with an inlay hint for example), or just clarifying the Nushell docs.

FrancescElies:
I also got bit by this in the passt. While I agree it is confusing when comming from other shells, nushell actually dedicates an entire section to this topic inside [getting started](https://www.nushell.sh/book/thinking_in_nu.html#nushell-s-environment-is-scoped). What else could have done the nushell mantainers to avoid us being bit by this? 

I am fine with the description, maybe you can add the link above to the broot docs? The obvious disadvantage is that if nushell restructure their docs or moves them elsewhere the link needs to be updated. Either way if you decide to add it or not I am fine with it.

Canop:
You guys know nushell better than me. I wait for an explicit approval before merging.

FrancescElies:
From my side all good, you can merge

PR #883, opened 2024-05-28
make it possible to unbind keys

### LINKED ISSUE #632: Unbind built-in keybinds
I find some of the built-in keybinds counterintuitive, like `<C-w>`, for example. I expect it to delete the previous word, but by default it closes panels and the program. There are a few other default keybinds that I don't like and there should probably be an option to disable them.

Canop:
To remove a binding (without binding it to another verb), define a verb without any execution, for example

    {
        key: alt-i
    }

Fix #632

### DIFF
--- a/src/app/panel_state.rs
+++ b/src/app/panel_state.rs
@@ -112,7 +112,7 @@ pub trait PanelState {
             .unwrap_or(internal_exec.bang);
         Ok(match internal_exec.internal {
             Internal::apply_flags => {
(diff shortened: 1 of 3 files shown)


chtenb:
Note: this is only possible when the key is not used at all. When you want to unbind a key and later bind it to something else, this doesn't work. I tried doing this because I wanted the help page (`:?`) to give a clean overview of the keys that are in use. Currently it also shows the keys that have been overridden.

PR #891, opened 2024-06-26
Use XDG_RUNTIME_DIR when available

yerlaser:
XDG_RUNTIME_DIR is almost guaranteed to be living inside in-memory filesystem (tmpfs) which is much more suitable for such a create and drop temp files. On many systems /tmp is still on hard disk.


FrancescElies:
LGTM, shortly tested on a windows11 with nu 0.95.

@Canop if that's enough for you feel free to merge.

Canop:
Why exactly the `is-not-empty` ? To check existence of the directory ? To know if it's used ?

PR #892, opened 2024-06-28
fixed couple of broken links in the README.md : )

eddiehazel:


Canop:
Thanks.

It's quite a pain right now to maintain both website/index.html and the README. I should probably generate one from the other one...

eddiehazel:
doesn't seem insurmountable ( : 

a github action?

Canop:
Or just change my existing build script (see the `.sh`), which has the bonus of running locally and thus avoiding pushing broken commits just to see the result.

eddiehazel:
as yes that is way more neat 👌

PR #899, opened 2024-07-07
don't canonicalize paths on windows when focusing a path

### LINKED ISSUE #812: When creating a new pane under Windows, the path contains garbage
image

This happens with `panel_right`. This syntax looks like some code might be trying to referring to the device ID of the disk, but failing.

This bug was not present earlier this year. Subsequent operations in broot DO fail because the garbled path gets passed to powershell, rmdir, xcopy, etc.

Canop:
Fix paths being changed to UNC on Windows.

Fix #812


AeliusSaionji:
Confirmed fixed!

Issue #888, opened 2024-06-09
Print Path doesn't print to stdout on windows with Br function in powershell


Alfamari:
Print Path doesn't print to stdout on windows with powershell br function, I've also even tried 2>&1 which I've had success with other programs that had issues printing to stdout like ffmpeg. This isn't really an issue in and of itself though because I easily created a workaround, but it leads to other problems I explain later:
`{

	name: "Print to stdout"
	invocation: "stdout"
	external: "echo {file}"
	from_shell: true
}`

I'm using bat as an example because I think it best illustrates the issue, though the issue is universal.

![wezterm-gui_aCdnOuRt1K](https://github.com/Canop/broot/assets/132071593/3c74325b-cb34-4337-8548-4d88823d3cb6)

![wezterm-gui_lh4n4HZBD9](https://github.com/Canop/broot/assets/132071593/fe7535cd-0bcd-4a7f-9c87-dcab4054bffd)

The issue stems from a couple of problems with staging multiple files.
One: "only verbs returning to broot on end can be executed on a multi-selection".
This eliminates the workaround as a possibility because the docs state "`from_shell` isn't compatible with `leave_broot = false`"
The workaround also doesn't work without `from_shell = true`.
Now `print_path` is able to be executed on multiple staged files, it just cannot be piped anywhere after.

Two: The default file operation commands operate sequentially which is substantially slower by spawning a brand new cmd instance each time rather than either putting all files on one line with cmd /c or using an array in powershell to let the shells handle the batch operations because there isn't a {staged_items} variable to contain multiple items yet. This has caused me to want to pipe for a speed increase and also to try and workaround some of the current limitations with staged items.
Like mentioned:
https://github.com/Canop/broot/issues/695#issuecomment-1518698471
https://github.com/Canop/broot/issues/266#issuecomment-972113711

I'm pretty sure this bug used to also apply to stuff like --help but it seems that got fixed.
In the process of making this submission I found that it works properly with the broot executable, just not the br function.

![wezterm-gui_qBxiWwymqX](https://github.com/Canop/broot/assets/132071593/eb58dac3-da2d-4530-b0d9-82ba4f06c2ad)

Though the extra newline at the end can cause issues with pipes:

![wezterm-gui_64PN91wWdP](https://github.com/Canop/broot/assets/132071593/c6507df4-620d-4387-a670-d52510330843)

I'm using Pwsh Core 7.4.2 but the issue is also present on windows powershell 5.1.

[...]

PR #901, opened 2024-07-08
:write_stdout and :writeln_stdout internals

Canop:
The `:writeln_stdout` internal allows recreating the `:print_path` verb:

```
    {
        invocation: pp
        cmd: ":write_stdout {file};:quit"
    }
```

but it's more generic.

With the `:write_stdout`, you can for example write the path without the trailing newline:

```
    {
        invocation: wp
        cmd: ":writeln_stdout {file};:quit"
    }

```


AeliusSaionji:
These are the verbs I tested against
```hjson
      {
          invocation: "echo_stdout_quit"
          external: "echo {file}"
          from_shell: true
      }
      {
          invocation: "test_stdout_quit"
          cmd: ":write_stdout {file};:quit"
      }
      {
          invocation: "testln_stdout_quit"
          cmd: ":writeln_stdout {file};:quit"
      }
      {
          invocation: "test_stdout"
          cmd: ":write_stdout {file}"
      }
      {
          invocation: "testln_stdout"
          cmd: ":writeln_stdout {file}"
      }
```
---
I'm seeing a few oddities here.

In no particular order:

1. Should `write_stdout` and `writeln_stdout` be callable from within broot? They don't accomplish much outside of verb use.
2. Is this new? It's preventing use of @Alfamari 's previously working `echo {file}` from shell.
image
3. In the staging area, the four `test*` verbs don't trigger the above error, but they also simply aren't found if more than one file is staged.
image
4. Staging area would be preferred, since it allows the user to see what they have done, and allows for correcting mistakes. However, via `testln_stdout` it is possible to accumulate files on stdout. Unfortunately this smacks us right into the original problem from #888 : there's a newline at the end of the list. Note: I reverted my changes in br.ps1 to test this PR.
image
5. With `test_stdout_quit` it is possible to pass _one_ file to powershell.
image
But of course, multiple invocations of `test_stdout` won't help with further files.
image
I do see a workaround - calling `testln_stdout` for some files, and `test_stdout_quit` on the last file. I don't know that this qualifies as a solution though.
image
6. This might be an unrelated bug, but in the course of testing these verbs some dozen times I noticed that tab complete happily cycles through all four verbs with `test<tab>`, but `testln<tab>` autocompletes to `testln_stdout_quit` and will not cycle with the other match `testln_stdout`. 🤷 

---

In summary, `write_stdout` and `writeln_stdout` narrowly avoid solving the problem :)
- `write_stdout` only supports passing one file
- `writeln_stdout` always contains a problematic newline
- Neither are available in the staging area

As a final thought - in addition to text output, it might be worth supporting json output, too. That's a trendy thing to do these days, and would play quite nicely with object oriented shells like powershell. Or any shell could leverage something like `jq`.

Canop:
A lot of good point.
Also https://github.com/Canop/broot/issues/888#issuecomment-2224324287

Canop:
An approach might be to allow more settings to be carried by verb arguments, using the syntax I introduced at https://github.com/Canop/broot/blob/main/CHANGELOG.md#v091---2019-07-29

This could for example be `write_stdout {file:comma-separated};:quit` or `write_stdout {directory:json-array}` or  `write_stdout {directory:lines}`.

I'll be cautious though, as I don't want to be carried into introducing user-side complexity solving only one problem.

Canop:
I'm dropping this PR and will try another approach.

PR #906, opened 2024-07-17
Update nerdfont and vscode icons

jpaju:
Added icon mappings for several file and extension types. Also fixed formatting and sorting on some mapping files


Canop:
I'm in holidays. Can I have a reviewer ?

FrancescElies:
nerdfont LGTM
grafik

vscode I have some issue together with wezterm, thus I didn't test it but that is probably related with my setup.

jpaju:
The icon for `a.Dockerfile` and `Dockerfile` actually seem to be off, it should be the same as `docker-compose.yml` 🤔

FrancescElies:
@jpaju not sure what's going on, does is look right on your side?

Even though a bit small I do get a funny tiny docker whale whale when printing the unicode char defined for it `( "file_type_docker", 0xf308 )`
 
grafik

FrancescElies:
@jpaju maybe you could move the docker maps to another pr? This way probably you could get this one merged and the docker icon problem can be looked in isolation there.

jpaju:
Yes! Now the icons for docker files are removed from this PR

FrancescElies:
@Canop if you are with this too feel free to merge.

PR #908, opened 2024-07-20
Update docker icons

jpaju:
Follow up for #906. Icons for `Dockerfile` and `a.Dockerfile` don't show up correctly


jpaju:
I'm unable to build broot locally so investigating the issue with docker icons is impossible right now. I'm on apple silicon mac and when running `cargo build` or `cargo install` I get an error:

```
error: linking with `cc` failed: exit status: 1
  |
  = note: env [...]
  = note: ld: library not found for -liconv
          clang-16: error: linker command failed with exit code 1 (use -v to see invocation)

error: could not compile `libz-sys` (build script) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
  ```

Instructions for building broot from source only list required dependencies from Linux distros so it could be that I'm missing some dependency 🤷 @FrancescElies any ideas how to resolve this?

**Comment by FrancescElies** [comment]
@jpaju I am not aware of having installed that lib explicitely on mac, does running `xcode-select --install` which installs apples commandlinetools help?. 
Otherwise maybe try to install it with `brew install libiconv`

**Comment by jpaju** [comment]
I'm using nix dev shell and issue was configuring that for mac.

<details>
<summary>If someone is interested, this flake.nix did it for me</summary>

```nix
{
  inputs.nixpkgs.url = "github:nixos/nixpkgs";
  inputs.flake-utils.url = "github:numtide/flake-utils";

  outputs = { nixpkgs, flake-utils, ... }:
    flake-utils.lib.eachDefaultSystem (system:
      let
        pkgs = import nixpkgs { inherit system; };
        darwinDeps = if pkgs.stdenv.isDarwin then with pkgs; [ libiconv darwin.apple_sdk.frameworks.Security ] else [ ];
      in {
        devShells.default = pkgs.mkShell {
          name = "broot";
          buildInputs = with pkgs; [ rustc rust-analyzer cargo openssl pkg-config ] ++ darwinDeps;
        };
      });
}
```

</details>

Now that I'm able to build locally, I was able to debug that the issue is in [font.rs](https://github.com/Canop/broot/blob/1f994018214d25753305f55fc1a1f5dbdbe1a5a6/src/icon/font.rs#L134-L136) `get_icon` function. Before passing `name`, `double_ext` and `ext` to `handle_file` function, they are lower cased. Because `Dockerfile` is capitalized, it fails to find icon.

I'm wondering whether it would be okay to remove `.to_ascii_lowercase()` from all three parameters when calling `handle_file` function?

PR #909, opened 2024-07-20
search_again internal

Canop:
This brings the `search_again` internal, which tries to fix a frequent annoyance, which is the loss of the search pattern when navigating, something I hadn't found good ergonomics for before.

`:search_again` is mapped by default to `ctrl-s`.

- When there's already an active filtering, and it wasn't complete, make it total (it thus does what the `total_search` internal was doing).
- When there's no active filtering, brings back the last used.

Main use case: you search, have a lot of matches, focus a folder to look inside, then you can bring back the search pattern with ctrl-s.

I'd be very interested in tester feedback on this.

### DIFF
--- a/src/app/app.rs
+++ b/src/app/app.rs
@@ -435,6 +435,29 @@ impl App {
                                 .refresh_input_status(app_state, &app_cmd_context);
                         }
                     }
+                    Internal::search_again => {
+                        if let Some(raw_pattern) = &self.panel().last_raw_pattern {
+                            let sequence = Sequence::new_single(raw_pattern.clone());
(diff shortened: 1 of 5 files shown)


FrancescElies:
Shortly tried on a mac, it felt good to be able to bring back the previous search with `ctrl-s`.

https://github.com/user-attachments/assets/bbd51754-5472-4ccd-b413-3fad80186161

Not sure if it's a better binding, but did you consider `ctrl-r` as a default? This would remind me of the reverse search in the shell. 
On the other hand I find the double behaviour of `ctrl-s` for full search and bring back last search pretty good too and makes me happy the way it is.

PR #913, opened 2024-07-23
Make sure temp file is always removed in Powershell launcher

stinos:
The try/finally will catch all possible errors.


Canop:
I have no idea of how those powershell things work, so, as usual, I'd like somebody else to confirm this PR is OK

Issue #818, opened 2024-01-10
Character inputs that require the Alt key are not working


mp-t:
When trying to input text to search for that requires usage of the Alt key, e.g. the backslash '\' on the German keyboard layout, broot just eats the input and does not register a character.

Tested on Windows Terminal and Windows conhost.
Broot version 1.26.1

Canop:
Many shortcuts involving the alt keys aren't usable in ANSI terminals.

You can use print_keys to check what's available: https://github.com/Canop/print_key

BTW I'm currently working at making those shortcuts available on the few terminals (eg Kitty, wezterm) which go beyond the ANSI capabilities.

Canop:
Many shortcuts involving the alt keys aren't usable in ANSI terminals.

You can use print_keys to check what's available, and how it's received by the TUI application: https://github.com/Canop/print_key

Canop:
I don't have the same keyboard than you so it's a little hard to be sure.
Can you tell me what print_key says when you type the intended combination, and whether it looks right to you ?

mp-t:
The right Alt key is interpreted as Ctrl+Alt.
This is reflected in the print_key output:

> You typed Ctrl-Alt-\

This looks correct to me.

FrancescElies:
I could maybe help with this one, i will give it a try the coming week

FrancescElies:
@Canop I could reproduce the issue. this personally does not affect me but since I happen to have german keyboard, for this experiment I switched from US to DE input.

Bottom right you see keys as I press them, where you see `ctr+alt+[` imagine `altgr+\`, don't ask me why the keyboard logger shows that instead of `altgr+\`.

![image](https://github.com/user-attachments/assets/58c13c3f-2cc2-4c96-82e6-b3a9ce40314d)


Below a screen recording of the thing.

https://github.com/user-attachments/assets/867b5064-3128-4477-b4a1-50e05435bc9b

@Canop shouldn't print-key see `\` instead of `Ctrl-Alt-\`?

Canop:
Is the problem specifically related to `ctrl-alt-\` or to any combination involving `alt-\` ?

FrancescElies:
any combination with altgr, same issue for the `}` and `]` which are on the left side of the `\`

![image](https://github.com/user-attachments/assets/d1a2be1b-d44f-408d-a13c-fed27b5a5414)

FrancescElies:
[...]

PR #921, opened 2024-07-31
Internals for changing the widths of panels

Canop:
3 new internals:

* `set_panel_width`, taking as parameter the index of the panel and the desired width
* `move_panel_divider`, taking as parameter the index of the divider and the desired change
* `default_layout`, restores the default panel sizes

`ctrl-<` is bound by default to `:move_panel_divider 0 -1`

`ctrl->` is bound by default to `:move_panel_divider 0 1`

It's also possible to define the layout in conf, eg 

```
    layout_instructions: [
         { panel: 0, width: 18 }
     ]
```


FrancescElies:
Shortly tested on windows with wezterm using US keyboard input.

image

I was not able to resize panes, on the right side you the key strokes I tried `< (shift+,)` `> (shift+.)`  `ctrl+< ((ctrl+shift+,))`and  `ctrl+> (ctrl+shift+,)`

This issue I am seeing might be related to #818

FrancescElies:
It's funny, pressing `<` and `>` in combination with `ctrl-shift` and `alt-shift`.

Print key shows the following.
image

At any rate, since this option is configurable and the issue seems to be with windows and one of the broo deps, the following bindings is a good workaround under windows.
```
    {
        invocation: move_divider_left
        key: ctrl-shift-,
        execution: ":move_panel_divider 0 -1"
        leave_broot: false
    }
    {
        invocation: move_divider_right
        key: ctrl-shift-.
        execution: ":move_panel_divider 0 1"
        leave_broot: false
    }
```

Unless you need any further testing from my side I will leave it here.

Canop:
Those comma and dot bindings aren't easy to type either. I still take suggestions.

FrancescElies:
They are for me on the same place as `<` `>`.

image

Honestly I don't have any good proposal.

PR #930, opened 2024-08-11
nushell icon

FrancescElies:
make it's icon an elepant.

grafik

Not sure if the elephant is the official logo but it's in the std lib, [here](https://github.com/nushell/nushell/blob/983014cc40e6a96548c731b7de26e3a1a70d6426/crates/nu-std/std/mod.nu#L185)
grafik

### DIFF
--- a/resources/icons/nerdfont/data/icon_name_to_icon_code_point_map.rs
+++ b/resources/icons/nerdfont/data/icon_name_to_icon_code_point_map.rs
@@ -522,7 +522,7 @@
 ( "file_type_nuget", 0xf15b ),
 ( "file_type_numpy", 0xf15b ),
 ( "file_type_nunjucks", 0xe679 ), // 
-( "file_type_nushell", 0xe691 ), // 
+( "file_type_nushell", 0xf07c6 ), // 󰟆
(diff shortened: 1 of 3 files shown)


Canop:
_I still don't have a list of the people here who follow nushell._

Opinion on that icon @texastoland @LudoPinelli @melMass @@stevenxxiu @mediumrarez ?

melMass:
Let's tag maintainers of nushell @fdncred @devyn @sholderbach

texastoland:
I think an official icon is still up in the air. I like the elephant and identify it with Nu too. The VS Code extension uses https://github.com/nushell/vscode-nushell-lang/blob/main/assets/nu.svg (just `nu`) however.

Canop:
Note that this can be reversed, so if nobody objects in the coming hours, we may adopt his icon.

*(and pretend PHP is dead)*

stevenxxiu:
In <https://github.com/nushell/nushell/tree/0.96.1/assets> it's the letters `nu>` instead of an elephant, so I'm not sure what's the better icon. Probably better for a maintainer to comment on this.

PR #931, opened 2024-08-13
Allow ignore/unignore rules in .ignore

### LINKED ISSUE #613: Ignore files without using .gitignore (.brootignore? .ignore?)
I'd like a way to tell broot to ignore certain folders with a large number of useless files (`~/.cache, ~/rust, ~/crates.io-index`) without making a git repo around them (or hiding them). Is there a way to do this already? If not, I suggest a .brootignore file, which is parsed the same as a .gitignore file, except it's not required to be in a git repo.

Canop:
Files called `.ignore` with the syntax of gitignore files are now read and used just like `.gitignore` files.

When, at the same level, broot finds rules from both a `.ignore` and a `.gitignore` files, the `.ignore` is the decisive one (it's assumed to be personal and meant to override the shared `.gitignore` file.

Eg I can have a global `.gitignore` file with 

    dys-notes.*
    
which helps me have notes everywhere and not sharing them.

And I can have a `.ignore` file, which reveals some of those files in broot:

    !dys-notes.md 
    
Fix #613

### DIFF
--- a/src/git/ignore.rs
+++ b/src/git/ignore.rs
@@ -168,7 +168,7 @@ impl GitIgnorer {
                     chain.push(self.files.alloc(gif));
                 }
             }
-            for filename in [".gitignore", ".git/info/exclude"] {
+            for filename in [".gitignore", ".git/info/exclude", ".ignore"] {
                 let file = dir.join(filename);
                 if let Ok(gif) = GitIgnoreFile::new(&file, dir) {
                     //debug!("pushing GIF {:#?}", &gif);
@@ -203,9 +203,11 @@ impl GitIgnorer {
             parent_chain.clone()
         };
         if chain.in_repo {
-            let ignore_file = dir.join(".gitignore");
-            if let Ok(gif) = GitIgnoreFile::new(&ignore_file, dir) {
-                chain.push(self.files.alloc(gif));
+            for filename in [".gitignore", ".ignore"] {
+                let ignore_file = dir.join(filename);
+                if let Ok(gif) = GitIgnoreFile::new(&ignore_file, dir) {
+                    chain.push(self.files.alloc(gif));
+                }
             }
         }
         chain


FrancescElies:
# tests


1. ✅ broot_*zip. 
Image: _upper-left (fd), bottom-left: broot, upper-right: .ignore, bottom-right: .gitignore_ grafik
 
1. ✅ Here I was expecting the zip file `releases/broot_/broot_.zip` to appear in broot but then I realized `fd` was behaving the same way, I had wrong expectations. 
Image: _upper-left (fd), bottom-left: broot, upper-right: .ignore, bottom-right: .gitignore_ grafik

1. ✅I did the same test as you described above. 
Image: _upper-left (fd), bottom-left: broot, upper-right: .ignore, bottom-right: .gitignore_ grafik

Canop:
Thanks for the tests and detailed report!

Regarding the 2: it's because when you ignore `releases`, you don't even have a look at what's inside.

Canop:
In the current implementation, when you're inside a git repository, `.ignore` files outside this repository aren't taken into account.

This mimics the behavior which was defined for git, but there are use cases which are thus not supported.

FrancescElies:
For the time being it’s ok, you can add those afterwards when there is enough people begging for those missing features. 

By the way it was so nice to see that being added to broot with such a small patch 🤩

Canop:
I'm currently adding those feature, after reflexion.

I want to support the case of having dys-notes.* always gitnored and always unignored, with only global files.

Canop:
@FrancescElies I made https://github.com/Canop/broot/pull/932

(I'll test it more but your opinion would of course be welcome)

Canop:
Merged. Documentation has been improved: https://dystroy.org/broot/tree_view/#hidden-gitignored-files

PR #932, opened 2024-08-14
Non erased global ignore

Canop:
The list of ignore files applicable to a directory contains
- the global `.gitignore` file (if any)
- all the `.ignore` files found in the current directory and in parents
- the `.git/info/exclude` file of the current git repository
- all the `.gitignore` files found in the current directory and in parents but not outside  the current git repository (i.e. not in git repositories containing the current git repository)

Deeper file have a bigger priority.
`.ignore` files have a bigger priority than `.gitignore` files.

---

Those rules allow for example having personal `my-notes.*` files always ignored by git but visible in broot, with only global files:

In `~/.config/git/ignore` :

```
.ignore
my-notes.*
```

In `~/.ignore` :

```
!my-notes.*
```

(note the `!` before the pattern)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ deploy.sh
 /press
 /glassbench_*.db
 .bacon-locations
+.ignore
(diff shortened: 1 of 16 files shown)


FrancescElies:
works nicely, thanks!

grafik


Over the next week I will be testing this in a windows setup too.

Canop:
> Over the next week I will be testing this in a windows setup too.

Thanks. So I'll wait for a few days before merging it (and releasing a new version of broot).

Issue #940, opened 2024-09-08
Termux: Termimad Error : IO error: Permission denied (os error 13)


dilongx:
Using broot version after 1.41.1 in Termux, I get "Termimad Error : IO error: Permission denied (os error 13)" May I ask why this happens and how to solve it?

Canop:
I don't maintain anymore any solution to build the termux version.

May I ask you how you managed to build the recent versions of broot for termux ?

dilongx:
> I don't maintain anymore any solution to build the termux version.
> 
> May I ask you how you managed to build the recent versions of broot for termux ?

It's an official build.
https://github.com/termux/termux-packages/blob/master/packages/broot/build.sh

I'm not sure if that's the reason, I'm just looking for a solution, maybe I should ask in the termux-packages project?

dilongx:
@Canop Thanks for the reply, they fixed it!

Canop:
for reference: https://github.com/termux/termux-packages/issues/21424

PR #945, opened 2024-09-19
Update rustix from 0.38.34 to 0.38.37

fornwall:
This gets https://github.com/bytecodealliance/rustix/pull/1147, a fix that makes things work on Android again (see #940 - this is the patch Termux carries when building latest broot).


Canop:
I can't check it on Android, so I'll just check it doesn't seem to break anything else (unlikely) before merging.

PR #948, opened 2024-09-25
Fix IME popup position

xubaiwang:
This PR fix the problem that [IMEs (input method editors)](https://en.wikipedia.org/wiki/Input_method) popup at wrong position, by manually setting cursor position after rendering of panels.

IME compatibility is often considered a necessity for non-Latin, especially East Asian users.

The comparison below shows what this PR does:

<table>
<tr>
<th>Current</th>
<th>This PR</th>
</tr>

<tr>
<td>

:x: IME popup at the last char rendered, which is the bottom right corner in this example.

</td>
<td>

:heavy_check_mark: IME popup at the end of the active panel input.

</td>
</tr>

<tr>
<td>

20240926_02h46m12s_grim

</td>
<td>

20240926_02h46m52s_grim

20240926_03h46m27s_grim

</td>
</tr>
</table>

~This PR also add  a config option to use the [CJK unicode width](https://docs.rs/unicode-width/latest/unicode_width/#cjk-feature-flag).~
~This option `use_cjk_width` defaults to `false`, so it won't affect non-CJK (Latin) context.~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -260,7 +260,7 @@ dependencies = [
  "strict 0.1.4",
  "syntect-no-panic",
  "tempfile",
- "termimad 0.30.0 (git+https://github.com/Canop/termimad?rev=5c691657592fe7798ed32d6f9bb1b36d44c93f87)",
+ "termimad",
  "terminal-clipboard",
  "terminal-light",
  "toml",
@@ -362,7 +362,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2873923833694e839cce5b846311ba5316e24b6e3726f8106368ebeaf3beb0d5"
 dependencies = [
  "clap",
- "termimad 0.30.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "termimad",
  "terminal-light",
 ]
 
@@ -937,7 +937,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tempfile",
- "termimad 0.30.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "termimad",
  "thiserror",
 ]
 
@@ -2162,24 +2162,9 @@ dependencies = [
 
 [[package]]
 name = "termimad"
-version = "0.30.0"
+version = "0.30.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "920e7c4671e79f3d9df269da9c8edf0dbc580044fd727d3594f7bfba5eb6107a"
-dependencies = [
- "coolor",
- "crokey",
- "crossbeam",
- "lazy-regex",
- "minimad",
- "serde",
- "thiserror",
- "unicode-width",
-]
-
-[[package]]
-name = "termimad"
-version = "0.30.0"
-source = "git+https://github.com/Canop/termimad?rev=5c691657592fe7798ed32d6f9bb1b36d44c93f87#5c691657592fe7798ed32d6f9bb1b36d44c93f87"
+checksum = "22117210909e9dfff30a558f554c7fb3edb198ef614e7691386785fb7679677c"
 dependencies = [
  "coolor",
  "crokey",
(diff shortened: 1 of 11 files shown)


Canop:
I'm not familiar with the CJK thing. Are there cases when `use_cjk_width=true` is more correct ? Does it depends on your terminal's settings ?

xubaiwang:
> Are there cases when use_cjk_width=true is more correct ? 

Due to some historical reason in Unicode encoding, some certain characters from different scripting incorrectly share the same codepoint. 

The most obvious case is the `“”` left/right double quotation mark (U+201C, U+201D). In Latin context, these two chars are of single width, while in East Asian CJK context, the double width is used.

In modern terminal emulators, this problem is often addressed with a combination of the font fallback list and width setting.
If mismatched, the double-width font glyph may be truncated, and even concealed in TUI escape sequence drawing, which is very position sensitive. 

> Does the east-asian context depend on your terminal's settings ? 

Yes, it depends. Many terminal emulators (e.g. foot, iTerm2, ...) have their own config option for setting it.
However, they have different defaults and there is no (standard) way to retrieve this information from terminal.

> Why should width_cjk be used only for the input and not for the many other width computations in broot ?

Actually the ambiguous width problem should be considered each time `UnicodeWithStr::width()` is used. 

But this requires huge amount of code change. It also involves the `InputField` widget from external `termimad` crate, while I do not have enough time to investigate into it this week.

**Conclusion**: I suggest that I can just revert the config option commit, split this PR for IME position only (which is also used by non-CJK users, e.g. Vietnamese, Indic, ...), and postpone the tackling of ambiguous unicode width to further efforts.

Canop:
> Actually the ambiguous width problem should be considered each time UnicodeWithStr::width() is used.
But this requires huge amount of code change. It also involves the InputField widget from external termimad crate, while I do not have enough time to investigate into it this week.

Of course, and the code to change is mainly in other crates (that I maintain). But I'm asking for the future.

> I suggest that I can just revert the config option commit

Why ? Even if it's not complete and only partially improve the situation, we may keep it, and add precisions in the config's comment until a more general change is made.

xubaiwang:
Okay. I'll change the PR's title to include "ambigous width", and will continue the work when I have spare time in October.

Canop:
If you're OK with that, I'll try to merge the PR as is soon and the work on CJK widths will be tracked in a dedicated issue.

xubaiwang:
Sorry. I have just found a bug that cursor goes to wrong position when content width exceeds input area width. 

I think the calculation of cursor position can be better handled in `termimad::InputField` and be exposed as public function. I'll fix it in further commits after making changes to the termimad crate. Please do not merge it now.

Canop:
> I think the calculation of cursor position can be better handled in termimad::InputField and be exposed as public function.

Right

> I'll fix it in further commits after making changes to the termimad crate. Please do not merge it now.

If you want, I can do the change in termimad myself.

Canop:
The CJK thing shouldn't be included so hastily in termimad, as the width are used for everything, this would be a more systemic change which doesn't seem trivial to do correctly without breaking performances.

xubaiwang:
> The CJK thing shouldn't be included so hastily in termimad

Sure. Exposing cursor position function should be adequate now for IME cursor position. 

As far as I know, such function exists on `termimad::InputFieldContent::cursor_pos()`, but not on `termimad::InputField`, which is used by `broot::PanelInput`. 
That would be a easy fix by adding wrapper methods to `InputField` and `PanelInput`.

> which doesn't seem trivial to do correctly without breaking performances

By the way, though I have to admit that adding such feature requires a lot of refacor and is not of first priority, 
the performance regression won't be a big issue, since the `unicode-width` crate use exactly same complexity for `width()` and `width_cjk()`.

Canop:
> By the way, though I have to admit that adding CJK width feature requires a lot of refacor and is not of first priority,

Yes, let's move that part to another issue which would need a set of test conditions before any development.

Canop:
@xubaiwang I just published termimad 0.30.1

xubaiwang:
Thanks. I think this PR is now ready for review.

Canop:
This looks good.

PR #952, opened 2024-10-20
remove temp files used by kitty

Canop:
In order to display images with the Kitty protocol, broot creates temp files at the optimal dimensions and gives them to the terminal.

A problem with this approach is that those images may take a lot of space, especially if you're on linux and never reboot, as the `/tmp` folder is usually emptied on reboot.

With this commit:
- broot, on quitting, deletes all the temp files it created
- broot has a maximum number of temp files (defined with the `kept_kitty_temp_files` config entry) and removes the least recently used one when the max is reached

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
+version = "0.2.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
+
+[[package]]
 name = "android-tzdata"
 version = "0.1.1"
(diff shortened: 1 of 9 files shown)


kovidgoyal:
Just FYI the kitty protocol allows you to tell the terminal emulator to delete the file after it has read the image data.

Canop:
@kovidgoyal Yes but we don't want to: when scrolling over images you're very likely to need the same image again very soon

Canop:
BTW @kovidgoyal I'm impressed. What's your trick to be notified when kitty is mentionned ? I'd need this superpower too...

kovidgoyal:
github lets you search across their entire database for keywords. I have a custom script that does that for me at an interval and notifies of new issues/prs/discussions for topics I am interested in. But if all you are interested in is one or two topics you can just use the github web frontends search function manually, it works reasonably well and just takes a minute or two a day.

PR #961, opened 2024-12-18
Remove eval usage in fish launcher

kidonng:
As stated in [fish documentation on eval](https://fishshell.com/docs/current/cmds/eval.html):

> If the command does not need access to stdin, consider using [source](https://fishshell.com/docs/current/cmds/source.html) instead.

This has been tested manually on latest fish, though it should be compatible with any relevant version.

### DIFF
--- a/src/shell_install/fish.rs
+++ b/src/shell_install/fish.rs
@@ -35,9 +35,8 @@ const FISH_FUNC: &str = r"
 function br --wraps=broot
     set -l cmd_file (mktemp)
     if broot --outcmd $cmd_file $argv
-        read --local --null cmd < $cmd_file
+        source $cmd_file
         rm -f $cmd_file
-        eval $cmd
     else
         set -l code $status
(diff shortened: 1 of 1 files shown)


kidonng:
Code golfing a bit the launcher could even be:

```fish
function br --wraps=broot
    function _broot
        broot --outcmd $argv[1] $argv[2..] && source $argv[1]
    end

    _broot (: | psub) $argv
end
```

Albeit cryptic for those unfamiliar with fish 😅

Canop:
I'm sad no other fish user can check this

Canop:
A side effect of your change is that the temp file is removed only after the execution of the command

kidonng:
True but practically no difference. Even if `source` fails for some mysterious reason `rm` will continue to be run just as fine; if the shell blows up after `source` then we have much bigger problems than not getting to clean up a temporary file.

kidonng:
Thanks!

Issue #956, opened 2024-12-02
Fails to build with GCC 15 (dependency onig_sys)


creideiki:
I'm not sure if broot can actually do anything about this, but other packagers will hit it eventually once GCC 15 is released, so here's a heads up:

Gentoo has started experimenting with using prereleases of GCC 15 as the system C compiler. When that is the case, `onig_sys`, which is a dependency of broot, fails to build. Downstream bug: https://bugs.gentoo.org/943785

`onig_sys` contains a copy of the Oniguruma C source code, which is used by default. That copy is very old, and in particular does not include the fixes from kkos/oniguruma#312 for GCC 15 compatibility.

I've filed a bug (rust-onig/rust-onig#191) and asked for an updated snapshot in `onig_sys`, but that project does not seem to be particularly actively maintained.

What I did instead in Gentoo was to tell `onig_sys` to use the system copy of the Oniguruma library (in Gentoo this is release 6.9.9 + GCC 15 patches), by setting the environment variable `RUSTONIG_SYSTEM_LIBONIG=1`. This seems to work, but means we're using a newer Oniguruma than intended.

Would you be interested in building other packages the same way, so we'll at least all be testing with vaguely the same version of Oniguruma? Or do you have some other suggestion of how to handle this problem?

Canop:
I think we can just build without onig. I'm testing whether it has some impact (might impact the performances of code preview).

Canop:
@creideiki Is the problem solved with https://github.com/Canop/broot/pull/957 according to you ?

creideiki:
Seems to work, with my limited testing - #957 builds with GCC 15 and still gives me syntax highlighting in source code.

Canop:
I didn't remove syntect, just compiled it so that it uses another regex lib, a pure rust one. So it probably works ok but there *might* be subtle bugs...

PR #968, opened 2024-12-30
Use a new version of the forked syntect

### LINKED ISSUE #967: Panic when previewing seemingly innocent JavaScript text file
broot 1.44.3 via Homebrew

```shellsession
$ echo 'export default new Map();' > test.js
$ broot test.js
thread 'main' panicked at /Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/index.crates.io-6f17d22bba15001f/syntect-no-panic-4.6.1/src/parsing/regex.rs:69:53:
regex string should be pre-tested: InvalidEscape("\\g")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Canop:
This fork doesn't panic on non compiling regexes.

Fix #967

The reason the fork is needed is that syntect panics when a regex from a syntax can't compile, which happens with sublime syntaxes when the engine isn't onig but fancy-regex.

The default regex engine of syntect is onig, but this unmaintained engine isn't pure-rust and isn't compatible with gcc 15, which limits compatibility for broot. So, right now, this fork is useful to prevent crashes.

The ideal solution would be either to have syntect completely handle regex compiling errors without panic (which looks easily doable but requires some work and may not appear as the best behavior for some other crates using syntect) and|or to have fancy-regex support all the regex features that are required by sublime syntaxes.

See also:
* https://github.com/fancy-regex/fancy-regex/issues/74
* https://github.com/sharkdp/bat/issues/3156 
* https://github.com/Canop/broot/issues/956
* https://github.com/trishume/syntect/issues/508


NoisyCoil:
Hi @Canop. Is the creation of a new fork (`syntect-5-no-panic`) really necessary? Can't the same be achieved by uploading a new version of the old fork (`syntect-no-panic`)?

In terms of packaging for distros, introducing a new dependency can take a lot of time (between weeks and months) and effort, while a simple update can usually be dealt with instantly.

Canop:
I didn't think this would be a problem. I'll revert to the old name.

Canop:
@NoisyCoil Reverted: https://github.com/Canop/broot/releases/tag/v1.44.5

NoisyCoil:
@Canop Thanks a lot!

NoisyCoil:
@Canop Could you please publish 1.44.5 to crates.io too?

Canop:
@NoisyCoil done

NoisyCoil:
@Canop Thanks!

PR #973, opened 2025-01-10
Add --server --force flag.

### LINKED ISSUE #969: Add a way to start a server instance with an existing instance name / socket
I am using the --listen function of broot together with zellij as an IDE together with helix. Closing the zellij session leaves the socket file behind. This has the side effect that a new server instance of broot with the same name can't be started:
```
➜~ br --listen br_server
Can't open socket: /tmp/broot-server-br_server.sock already exists - consider removing it
```

It would be convenient to have a flag for forcefully delete the sock file in case there is already there, e.g.
` br --listen br_server --force`

VasilisManol:
* Add a --force flag for server in the case the socket already exists and you need to forcefully
delete the file. Fixes #969.

### DIFF
--- a/src/net/server.rs
+++ b/src/net/server.rs
@@ -27,7 +27,10 @@ impl Server {
     ) -> Result<Self, NetError> {
         let path = super::socket_file_path(name);
         if fs::metadata(&path).is_ok() {
-            return Err(NetError::SocketNotAvailable { path });
+            match fs::remove_file(&path) {
+                Ok(_) => {}
+                Err(e) => return Err(NetError::Io { source: e }),
+            }
         }
         let listener = UnixListener::bind(&path)?;
         info!("listening on {}", &path);


Canop:
I have no problem with the feature but what bothers me here is the addition of a launch argument.

Anybody here can comment ? Any alternative to a new argument ? Any idea for making its name more obvious ?

VasilisManol:
> I have no problem with the feature but what bothers me here is the addition of a launch argument.
> 
> Anybody here can comment ? Any alternative to a new argument ? Any idea for making its name more obvious ?

For my use case, I would actually prefer it to have it by default. I added the flag just for not changing the existing behavior.

Canop:
Given that nobody commented, I suggest you change the PR by removing the additional argument, making it delete the file by default, and we'll see what we do if somebody files an issue for regression.

Canop:
Damn... I saw no change to the PR but refreshed the page by luck. Sorry to make you wait 2 weeks. I'll handle it soon.

VasilisManol:
No worries :)

PR #1002, opened 2025-04-20
Fix missing `--max-depth` argument in nushell install

lizclipse:
Without this you get errors upon loading:

```nu
Error: nu::parser::variable_not_found

  × Variable not found.
    ╭─[/Users/liz/.config/nushell/autoload/br.nu:76:8]
 75 │     if $show_root_fs { $args = ($args | append $'--show-root-fs') }
 76 │     if $max_depth != null { $args = ($args | append $'--max-depth=($max_depth)') }
    ·        ─────┬────
    ·             ╰── variable not found.
 77 │     if $sort_by_count { $args = ($args | append $'--sort-by-count') }
    ╰────
```


Canop:
Thanks. I'll release a new version with this fix.

PR #1004, opened 2025-04-21
Add type hint to `max-depth` for nushell launcher

sandyspiers:
The nushell launcher is currently broken:

```
> br
error: invalid value 'false' for '--max-depth <MAX_DEPTH>': invalid digit found in string
```

The `max-depth` argument needs to be type-hinted to integer, so that if its blank it defaults to `null`, as opposed to `false`.

Thanks for the awesome program, really enjoying it!


Canop:
@amitkot @sandyspiers Are you confident I can release that without trying it on nushell myself ?

sandyspiers:
Its working for me locally in nushell

amitkot:
It worked for me as well.

BTW, I would also add to the documentation the requirement to source this file as part of my config.nu, as this was not added automatically via the `broot --install` call.

Canop:
> BTW, I would also add

PR welcome

PR #1026, opened 2025-06-24
display texts with control chars as texts 

### LINKED ISSUE #977: Allow text file preview for all UTF-8 encoded files
A number of PDF files converted into text (`mutool convert...`) still have few innocuous control characters (typically 0x0c replacing a page break). But these bug Broot into displaying them as hex dumps.

Please consider this patch below for Broot to be more lenient and accept all UTF-8 encoded files, simply by replacing control characters with space (or an empty string).

Binary files are unaffected by this patch, as they get rejected earlier by not being UTF-8 compliant.

```rs
diff --git a/src/syntactic/syntactic_view.rs b/src/syntactic/syntactic_view.rs
             self.total_lines_count += 1;
             let start = offset;
             offset += line.len();
+            line = line.replace(|ch: char| ch.is_control() && !"\t\n\r".contains(ch), " ");
-            for c in line.chars() {
-                if !is_char_printable(c) {
-                    debug!("unprintable char: {:?}", c);
-                    return Err(ProgramError::UnprintableFile);
-                }
-            }
```

Canop:
Some of those control chars (the ones that seem to break TTY rendering in my small test) are replaced with '�'.

Fix #977

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -346,6 +346,7 @@ dependencies = [
  "umask",
  "unicode-width",
  "uzers",
+ "vte",
  "which",
  "xterm-query",
 ]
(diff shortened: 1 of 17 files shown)


paul-d-ray:
I can confirm this works.  

I tested about 25 different pdf files that would display as hex before this fix.  

I am using pdftotext.

```
    {  
        input_extensions: [ "pdf" ]  
        output_extension: txt  
        mode: text  
        command: [ "pdftotext", "-layout", "-enc", "UTF-8", "{input-path}", "-" ]  
    }
```

Canop:
released in broot 1.47.0

PR #1050, opened 2025-09-16
revert: don't use Unicode placeholders by default for the *Kitty* terminal emulator, as this is for *Tmux*

stevenxxiu:
I just realized Unicode placeholders should only be used in *Tmux*, so using it by default isn't quite right.


stevenxxiu:
Not much really. There's a few more bytes transmitted for the placeholder grid.

It's more due to the spec [Terminal graphics protocol - kitty](https://sw.kovidgoyal.net/kitty/graphics-protocol/#unicode-placeholders) saying:

> This approach is less flexible, but it allows using images inside any host application that supports Unicode, foreground colors (tmux, vim, weechat, etc.), and a way to pass escape codes through to the underlying terminal.

PR #1066, opened 2025-10-02
fix non-trash platforms

dead10ck:
279a4af removed the `trash` feature in favor of just using conditional compilation for supported platforms; but it didn't actually compile on platforms not supported by the `trash` crate.

This fixes compilation in these cases. Rather than compile out the trash command altogether, this prints an error message if trash is not supported on the platform. This way, key bindings and such can still work across platforms.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,6 @@ termimad = "0.34"
 terminal-clipboard = { version = "0.4.1", optional = true }
 terminal-light = "1.8"
 toml = "0.9"
-trash = "5.2"
 umask = "2.1.0"
 unicode-width = "0.2"
 vte = "0.15"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -74,6 +73,9 @@ xterm-query = { version = "0.5", optional = true }
 [dev-dependencies]
 glassbench = "0.4.4"
 
+[target.'cfg(any(target_os = "windows", all(unix, not(any(target_os = "macos", target_os = "ios", target_os = "android")))))'.dependencies]
```
Canop, line 76: But the trash feature compiles and works on MacOS

Canop (changes requested):

Canop:
On what platform do you have a trash related compilation problem without this PR ?

Cargo.toml
dead10ck: Oh ok, I had just copied the same conditional guard that was used here: 

https://github.com/Canop/broot/blob/3de3c2e44c536df645a8b1b81935b611ad667577/src/verb/verb_store.rs#L173-L181

Should these extra trash commands be available on MacOS too?

Canop:
Merged, thanks

PR #1070, opened 2025-10-02
use zune-image for improved speed

Canop:
Use the zune-image crate instead of the image one, unless when zune-image doesn't work (gif, tiff, webp, etc.).

This improves the rendering speed of image preview.


Canop:
Tests welcome, especially as the improvement may be cpu dependant (zune uses SIMD optimizations).

PR #1072, opened 2025-10-04
Completion back tab

### LINKED ISSUE #1071: Go back during verb completion?
How does one go one step back during verb completion? `shift-Tab` does not seem to be working. (Tested 1.49.1 and 1.46.4 and tried it with both WezTerm and Alacritty).

When I type `:togg` and then start pressing `Tab` I can browse all verbs having prefix `togg` in some order but how do I browse them backwards?

Manual somewhat hints at `next_match` and `previous_match` keybindings being used during autocompletion with default binding being

```
previous_match│        │shift-BackTab, BackTab│select the previous match
```

But I don't know what the BackTab key is.

Adding
```
{
    key: shift-Tab
    internal: previous_match
}
```
to verbs has no effect. I also tried setting it to `ctrl-x` and `alt-x` just in case `shift-Tab` is handled differently e.g. being eaten by terminal etc.. but nothing seems to work.

I hit this problem within seconds of trying to use Broot (after many years of not using it) since browsing **back** and forth in commands autocompletion is a natural way for me to explore available commands.

Am I doing something wrong or is it just a bug? If it is just a bug perhaps I can look at fixing it.

Canop:
Fix #1071

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - improved image rendering (both speed by using the zune-image library, and quality with bilinear interpolation)
 - fix compilation broken by 1.50.0 on Android- thanks @dead10ck
 - `--listen-auto` listens for commands on a random linux socket - Fix #1064
+- when auto-completing, `back-tab` cycles in reverse order - Fix #1071
 
 ### v1.50.0 - 2025-09-25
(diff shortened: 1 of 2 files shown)


WizardUli:
Works as expected!

Issue #1082, opened 2025-10-17
open editor at line number of first occurance if performing text/regex/composit search


JoshuaPostel:
Is there a way to have broot execute `$EDITOR +{line_number_of_first_occurcance} {file}` when opening a file after performing a text search i.e. `c/some_text_im_looking_for:e`?

Canop:
Thanks for spotting the documentation mismatch: the default edit verb is `$EDITOR {file}` because not all editors accept a line argument, and the doc still described the old `+{line}` form. Fixed. Closing.

PR #1083, opened 2025-10-17
Update verbs.hjson to match website documentation

JoshuaPostel:
See #1082 for discussion of website documentation not matching default configs.  This might not be an appropriate default as the `+{line}` argument will likely break many user's default $EDITOR


Canop:
After rereading the comments on the verb, I think the current solution is best: not adding by default an argument which isn't compatible with all editors, but explaining how to add it for the relevant editors.

PR #1096, opened 2025-11-07
Fix preview filter prefilling when negation is used

maxim-uvarov:
When using !cr/ (negated regex content filter), the preview filter should not be prefilled with the regex pattern. This fix adds a check in tree_to_preview() to return an empty pattern when the raw pattern starts with '!' (negation operator).

Fixes the issue where preview filter was incorrectly prefilled when using negated content regex filters like !cr/pattern/.

P.S. I know 0 of rust. Claude crafted this for me.


Canop:
Yes

Canop:
I'll merge the other PR and try to release a new version of broot this week-end.

PR #1114, opened 2026-01-12
feat: detect iTerm for kitty terminal graphics protocol

### LINKED ISSUE #905: Support iTerm image protocol
It would be awesome if broot adds support to [iTerm image protocol](https://iterm2.com/documentation-images.html) in addition to existing kitty and WezTerm support.

Implementations in other programs, may be useful:

- [viuer](https://github.com/atanunq/viuer) (Rust)
- [timg](https://github.com/hzeller/timg) (C++)

I'd like help to test (or even contribute given some pointers) as I assume your main OS is Linux and may not have access to iTerm.

kidonng:
iTerm2 added kitty image support for some time but it has been buggy until recently: https://gitlab.com/gnachman/iterm2/-/issues/12483

Screenshot 2026-01-12 at 6 18 35 PM

Resolve #905

### DIFF
--- a/src/kitty/detect_support.rs
+++ b/src/kitty/detect_support.rs
@@ -61,6 +61,19 @@ pub fn detect_kitty_graphics_protocol_display() -> KittyGraphicsDisplay {
         } else if term_program == "ghostty" {
             debug!("Ghostty implements Kitty Graphics protocol");
             return KittyGraphicsDisplay::Direct;
+        } else if term_program == "iTerm.app" {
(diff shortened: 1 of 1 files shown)


stevenxxiu:
The config workaround sounds good. Checking for *iTerm2* explicitly and working around its bug seems less ideal. In the linked issue, the author was pretty quick to reply anyway, so it's probably easier to just raise a bug there.

kidonng:
Thanks for getting that edge case resolved!

PR #1116, opened 2026-01-19
fix: switch `$nu.temp-path` to `$nu.temp-dir`

stevenxxiu:
See <https://www.nushell.sh/blog/2026-01-17-nushell_v0_110_0.html#renamed-nu-temp-path-and-nu-home-path-toc>.

### DIFF
--- a/src/shell_install/nushell.rs
+++ b/src/shell_install/nushell.rs
@@ -124,7 +124,9 @@ export def --env br [
     if $write_default_conf != null { $args = ($args | append $'--write-default-conf=($write_default_conf)') }
 
     let cmd_file = (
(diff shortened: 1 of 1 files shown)


Canop:
How does it go with people having an older nushell installation ?

Canop:
Do you feel confident enough or do you want to wait for other nushell+broot users to give feedback ?

(I think GitHub is missing a system to define per-repo teams)

Canop:
> I wouldn't mind having someone else give feedback, but you probably know this almost never happens on GitHub.

Yeah, right, and doesn't happen much outside of GitHub too ^^

Canop:
The VERSION should probably be updated so that people having already the shell function installed have it reinstalled.

BTW, it looks like it's a pressing matter and I should have made a new release: https://github.com/Canop/broot/issues/1120

PR #1122, opened 2026-01-29
Climb path on missing working directory when restoring from terminal mode

mrnossiom:
Hi 👋🏻,

If you
- enter a directory in terminal mode
- delete this directory from somewhere else (e.g. a file explorer when tiding files)
- exit the terminal mode in the deleted directory

broot crashes while not having completely restored terminal capabilities.

This PR fixes this issue by trying to climb the path up.

e.g. you were in `/tmp/folder/i-should-delete-this/`, when the folder is deleted, br's cwd is set to `/tmp/folder` instead of crashing.

I restricted the error to `ErrorKind::NotFound` but maybe it should include `ErrorKind::PermissionDenied` too.

Thanks!

### DIFF
--- a/src/launchable.rs
+++ b/src/launchable.rs
@@ -219,12 +219,15 @@ impl Launchable {
                             break;
                         };
 
-                        if err.kind() == std::io::ErrorKind::NotFound && dir.parent().is_some() {
-                            dir = dir.parent().unwrap();
-                            continue;
+                        if err.kind() != io::ErrorKind::NotFound {
+                            return Err(ProgramError::Io { source: err });
                         }
 
-                        panic!("Unable to restore the working directory: {}", err);
+                        let Some(parent_dir) = dir.parent() else {
+                            return Err(ProgramError::Internal { details: "could not restore from terminal mode".to_string() });
+                        }
+
+                        dir = parent_dir;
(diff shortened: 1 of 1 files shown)


src/launchable.rs
```diff
@@ -211,7 +211,21 @@ impl Launchable {
...
+                        let Err(err) = std::env::set_current_dir(dir) else {
+                            break;
+                        };
+
+                        if err.kind() == std::io::ErrorKind::NotFound && dir.parent().is_some() {
+                            dir = dir.parent().unwrap();
+                            continue;
+                        }
+
+                        panic!("Unable to restore the working directory: {}", err);
```
Canop: The original unwrap was lazy and bad.
I should have written something like

                    std::env::set_current_dir(old_working_dir)
                        .map_err(|source| ProgramError::Io { source })?;

Going up until it works is a good idea for a better experience, but please don't put other `unwrap` and `panic` in the code. Use a pattern like `if let Some` to avoid the first unwrap and return an error instead of panicking when the loop doesn't solve the problem.

Canop:
I'll fix the error type when I merge (which will be in a few days as I'm deep in a change involving quite a refactor)

PR #1124, opened 2026-02-08
impacted_panel verb parameter

### LINKED ISSUE #1119: Request for new verbs: variants of (page_up, page_down) for the right-most pane, without changing focus
Hi, and thanks as always.

I'd like to be able to scroll or page a preview pane while keeping a left-pane file tree focused.

I like the following behavior of pgup/pgdown in [tig](https://github.com/jonas/tig):

- with a single pane (git log lines), page the open pane
- with two panes (git log lines, git diff view), page the right pane

Canop:
This adds an optional `impacted_panel` parameter to verbs, allowing a command to apply to another panel than the one in which it was triggered.

Example:

    {
        key: alt-pagedown
        internal: page_down
        impacted_panel: right
    }
    {
        key: alt-pageup
        internal: page_up
        panel: right
    }

The action of these verbs always apply to the rightest panel. This allows for example scrolling the preview panel without focusing it.

Fix #1119 

Right now, the possible values of impacted_panel are:

* `active` (default) 
* `left(est)`
* `right(est)`
* `id:<positive-integer>`
* `idx:<positive-interger>`
* `preview`

This is experimental, feedback welcome

### DIFF
--- a/src/app/app.rs
+++ b/src/app/app.rs
@@ -178,8 +178,8 @@ impl App {
                     let screen = self.panels.screen();
                     self.panels.refresh_active_panel(con);
                     if let Some(new_arg) = new_arg {
-                        self.panels.mut_panel().set_input_arg(new_arg);
-                        let new_input = self.panels.panel().get_input_content();
+                        self.panels.set_input_arg(new_arg);
+                        let new_input = self.panels.get_input_content();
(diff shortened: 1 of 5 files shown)


Canop:
As there's a big refactoring with this PR, I'd appreciate people tests

PR #1140, opened 2026-03-09
space-separated/comma-separated verb arg flags

### LINKED ISSUE #465: Make it possible for a verb to issue a single command for the whole list of staged files
Primary mentioned here: https://github.com/Canop/broot/issues/266#issuecomment-972113711

It should be possible to execute a command with argument the space separated list of paths

Canop:
This PR adds the `space-separated` and `comma-separated` flags for verb arguments.

With none of those flags, a verb called on a multiple selection (i.e. the stage) results in one command per selected file.

When there's one of those flags, only one command is generated, merging all selections.

Example:

```Hjson
    {
        invocation: "zip {name}"
        external: [
            "zip"
            "-r"
            "{name:path-from-directory}.zip"
            "{file:space-separated}"
        ]
        leave_broot: false
        working_dir: "{root}"
    }

``` 

As there's one of the two new flags, when several files are selected, a unique command is executed, eg

    zip -r my-new-archive.zip first/path second/path.rs
    
 Fix #465

### DIFF
--- a/resources/default-conf/verbs.hjson
+++ b/resources/default-conf/verbs.hjson
@@ -22,14 +22,14 @@ verbs: [
     #  so that you may jump directly at the right line from a preview or
     #  a content search.
     # Examples depending on your favourite editor:
-    #    execution: "nvim +{line} {file}"
-    #    execution: "helix {file}:{line}"
+    #    external: "nvim +{line} {file:space-separated}"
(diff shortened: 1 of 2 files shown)


SplendidSwifter:
Documentation at [Single command on stage](https://dystroy.org/broot/conf_verbs/#single-command-on-stage)

PR #1157, opened 2026-04-23
Feature: add favorite panel

shejingui:
Summary of latest commit:

- Commit: 51b03c677a086b0d25e69557b94c2d4af8da4bd6
- Author: jingui.she
- Date: Wed Apr 22 19:42:17 2026 -0700
- Feature: add favorite panel

Broot is a great program, and I use it every day. 
Inspired by the stage panel, I would like to added a new “Favorites” panel. 
This panel allows you to bookmark any file, and its usage is almost the same as the stage panel. 
when it's the rightmost panel, press ctrl-right can show the preview for selected item.
The items in the Favorites panel can be saved and automatically reloaded when Broot restarts.
I hope this feature helps with daily use.

I also fixed the following long-standing bugs:
1. When there are three panels (tree, stage, and preview), pressing Ctrl+Right would cause the tree panel to close.
2. If the leftmost tree panel is an ArgEdition panel and it is the only tree panel, pressing Ctrl+P could cause the panel to end and the program to exit unexpectedly.

usage of favorite panel
image

screen shot of favorite panel
image


angeluzmortem6891-glitch (changes requested):
😏☺️☺️

nnsite:
why not just allow for multiple, named stage areas ?
so you can stage for deletion, moving, symlinking or whatever in the same workflow without confusion.
as for what happens if files are in multiple places, leave it to the user.

shejingui:
> why not just allow for multiple, named stage areas ? so you can stage for deletion, moving, symlinking or whatever in the same workflow without confusion. as for what happens if files are in multiple places, leave it to the user.

The staging area is intended for temporary panels. Broot currently does not support persisting files, so this area is cleared every time you restart the program.

The Favorites panel is similar to the favorites or bookmarks feature found in file managers. It allows you to pin frequently accessed files so that you don't have to search for them every time. The selections are persisted across program restarts, which is a common feature in file managers such as Superfile.

These two features serve different purposes and use cases.

Canop:
Note: I'm late to review this, but I'm not dropping this PR, I'll come back to it.

PR #1158, opened 2026-04-27
Fix links in README to GitHub repositories

aripitek:
Updated links in README to point to GitHub repositories.


Canop:
User reported.

PR #1159, opened 2026-05-06
fix: iTerm version check

kidonng:
Follow-up to #1114. Previous check was copied from existing check for WezTerm, which didn't account for regular semver used by iTerm. This means current latest iTerm version (3.6.10) is considered incompatible.

### DIFF
--- a/src/kitty/detect_support.rs
+++ b/src/kitty/detect_support.rs
@@ -65,7 +65,7 @@ pub fn detect_kitty_graphics_protocol_display() -> KittyGraphicsDisplay {
             if let Ok(version) = env::var("TERM_PROGRAM_VERSION") {
                 debug!("$TERM_PROGRAM_VERSION = {version:?}");
 
-                if &*version < "3.6.6" {
+                if is_version_before(&version, "3.6.6") {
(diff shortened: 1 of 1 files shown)


src/kitty/detect_support.rs
```diff
@@ -148,3 +148,21 @@ pub fn is_ssh() -> bool {
     }
     false
 }
+
```
Canop, line 151: If we're going to have a version comparison function, we should probably have a semver compatible one, no ?

I think this function doesn't consider "1.2.0-beta" as before "1.2.0".

kidonng: I agree with this in principle, but I think current check is good enough unless it would be used more seriously. It was rare enough for people to stay on a old version for a piece of software like iTerm, let alone on a beta version.

Maybe a comment noting the incompleteness and explaining the rationale could do some help?

Canop:
Sorry, I forgot this PR. It's been made obsolete by a commit which landed in 1.59

Canop:
Added the missing credit @kidonng 
https://github.com/Canop/broot/commit/24b4c999b1ef0262cb8cc4b6bdd7a69e1c2c568b

PR #1167, opened 2026-05-30
fix: resolve #849 — Username with one or more spaces causes an error

### LINKED ISSUE #849: Username with one or more spaces causes an error
# Steps to reproduce
1. (optional) change your username to include one or more spaces
2. If you have br installed already, run `br --install`
3. run `broot --install`
4. the command above creates $HOME\Documents\WindowsPowershell\Profile.ps1 with the following contents
(when username is "User Name")
```powershell
. C:\Users\User\ Name\AppData\Roaming\dystroy\broot\config\launcher\powershell\br.ps1
```

# Possible solution
Use $HOME environment variable instead so windows will automatically expand it on execution appropriately.

chinhkrb113:
## Summary

The `broot --install` command writes a line into `Profile.ps1` that dot-sources `br.ps1` using a raw, unquoted absolute path:

Fixes #849

## Changes

- `src/shell_install/powershell.rs`

## Why

`src/shell_install/powershell.rs`: The `broot --install` command writes a line into `Profile.ps1` that dot-sources `br.ps1` using a raw, unquoted absolute path:

### DIFF
--- a/src/shell_install/powershell.rs
+++ b/src/shell_install/powershell.rs
@@ -24,7 +24,10 @@ use {
     directories::UserDirs,
     std::{
         fs,
-        path::PathBuf,
+        path::{
+            Path,
+            PathBuf,
+        },
         process::Command,
     },
     termimad::mad_print_inline,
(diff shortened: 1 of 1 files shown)


chinhkrb113:
Hi @Canop 
Do you still need any additional Windows confirmation before merging it? I’m happy to help verify if needed.

Canop:
I was waiting for confirmation indeed, because I don't know Windows much and I'm not able to test right now, but as nobody apart you looks interested enough, I think I won't delay the merge further.

Unless you think about something else, I'll merge in a few hours.

PR #1168, opened 2026-05-30
fix: resolve #963 — br error on Right/open of md associated editor (zsh/vscode)

### LINKED ISSUE #963: br crashes when exiting and then attempting re-enter of md associated editor (vscode)
```
 (node:25796) [DEP0168] DeprecationWarning: Uncaught N-API callback exception detected, please run node with option --force-node-api-uncaught-exceptions-policy=true to handle those exceptions properly.
(Use `Code --trace-deprecation ...` to show where the warning was created)
[25796:1219/174045.069:ERROR:jump_list.cc(300)] Failed to append custom category 'Recent Folders' to Jump List due to system privacy settings.
[main 2024-12-19T23:40:45.085Z] updateWindowsJumpList#setJumpList unexpected result: customCategoryAccessDeniedError
[main 2024-12-19T23:40:50.333Z] Extension host with pid 28008 exited with code: 0, signal: unknown.
```

environment is msys2 zsh via git-bash

chinhkrb113:
## Summary

When broot opens a file with an associated GUI editor (e.g. VSCode on Windows/msys2), the editor is launched as a background program (switch_terminal = false). broot currently spawns this child without redirecting its stdin/stdout/stderr, so any diagnostic output the editor writes (deprecation warnings, jump-list privacy errors, extension host messages) is written straight into broot's terminal and surfaces as the "error" the user sees in this issue. The fix is to fully detach background programs from broot's terminal by piping their standard streams to null. This keeps broot's TUI clean and matches the intent that background launches should not share broot's console

Fixes #963

## Changes

- `src/launchable.rs`

## Why

`src/launchable.rs`: When broot opens a file with an associated GUI editor (e.g. VSCode on Windows/msys2), the editor is launched as a background program (switch_terminal = false). broot currently spawns this child without redirecting its stdin/stdout/stderr, so any diagnostic output the editor writes (deprecation warnings, jump-list privacy errors, extension host messages) is written straight into broot's terminal and surfaces as the "error" the user sees in this issue. The fix is to fully detach background programs from broot's terminal by piping their standard streams to null. This keeps broot's TUI clean and matches the intent that background launches should not share broot's console.

### DIFF
--- a/src/launchable.rs
+++ b/src/launchable.rs
@@ -36,6 +36,7 @@ use {
         path::PathBuf,
         path::Path,
         process::Command,
+        process::Stdio,
     },
     which::which,
 };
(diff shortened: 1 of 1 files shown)


Canop:
Right-arrow on a file goes through `Internal::open_stay` → `opener::open` (browser_state.rs:163), not through `Launchable::Program`, so the patched branch isn't reached in that scenario. switch_terminal also doesn't mean "background program": it only controls the alternate/normal terminal switch, and nulling the streams there would silence the output of legitimate leave_broot verbs that set it to false.

More generally, discarding a launched program's stderr isn't a direction I want to take: when a verb fails, that output is what the user needs to see.

I'm closing this. As noted in the issue, the messages come from VSCode itself whatever launches it.

PR #1170, opened 2026-05-30
fix: resolve #788 — `br` command fails on Windows

### LINKED ISSUE #788: `br` command fails on Windows

chinhkrb113:
## Summary

The generated PowerShell `br` function passes the temporary command file path to broot via a single interpolated `-ArgumentList` string. When the path contains a space (common when the Windows username contains a space, e.g. `C:\Users\First Second\AppData\Local\Temp\...`), `Start-Process` splits the string on whitespace. broot then receives a truncated `--outcmd` value and the trailing path segment as an unexpected extra argument, producing the reported `File not found` error. The temp path must be quoted so it survives argument tokenization

Fixes #788

## Changes

- `src/shell_install/powershell.rs`

## Why

`src/shell_install/powershell.rs`: The generated PowerShell `br` function passes the temporary command file path to broot via a single interpolated `-ArgumentList` string. When the path contains a space (common when the Windows username contains a space, e.g. `C:\Users\First Second\AppData\Local\Temp\...`), `Start-Process` splits the string on whitespace. broot then receives a truncated `--outcmd` value and the trailing path segment as an unexpected extra argument, producing the reported `File not found` error. The temp path must be quoted so it survives argument tokenization.

### DIFF
--- a/src/shell_install/powershell.rs
+++ b/src/shell_install/powershell.rs
@@ -41,7 +41,7 @@ Function br {
 
   try {
     $process = Start-Process -FilePath 'broot.exe' `
-                            -ArgumentList "--outcmd $($cmd_file.FullName) $args" `
+                            -ArgumentList "--outcmd `"$($cmd_file.FullName)`" $args" `
                             -NoNewWindow -PassThru -WorkingDirectory $PWD
 
     Wait-Process -InputObject $process #Faster than Start-Process -Wait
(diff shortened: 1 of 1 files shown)


Canop:
Thanks.

I took the liberty to change your commit a little when dealing with the conflict. You're still the author of the commit: https://github.com/Canop/broot/commit/94158abc1cf2973b8e5ff9e59b7ccae00a7d2e44

Merged.

PR #1171, opened 2026-05-30
fix: resolve #509 — Image rendering not "full quality" on Wezterm

### LINKED ISSUE #509: Image rendering not "full quality" on Wezterm
I tried previewing `.png`s on Broot using Wezterm on both Windows and Arch Linux, and while Wezterm offer HD image previews with its imgcat function, Broot displays a poor image preview (blocky image preview).

At first I suspected it to be a Windows problem, but I wonder why it does not work as expected even on Linux.

On Windows (WezTerm)

Whereas WezTerm's imgcat is much sharper:
image

What am I missing, since the website mentions that Wezterm supports clear images [here](https://dystroy.org/broot/#preview-files)?

chinhkrb113:
## Summary

fix: resolve #509 — Image rendering not "full quality" on Wezterm

Fixes #509

## Changes

- `src/kitty/detect_support.rs`

## Why

`src/kitty/detect_support.rs`:

### DIFF
--- a/src/kitty/detect_support.rs
+++ b/src/kitty/detect_support.rs
@@ -4,6 +4,17 @@ use {
     std::env,
 };
 
+/// Whether the WezTerm build identified by `$TERM_PROGRAM_VERSION`
+/// supports the Kitty Graphics protocol.
+///
+/// A missing version is assumed to be a recent, supporting build.
+fn wezterm_supports_kitty_graphics(version: Option<&str>) -> bool {
+    match version {
+        Some(version) => version >= "20220105-201556-91a423da",
+        None => true,
+    }
+}
+
 /// Determine whether Kitty's graphics protocol is supported
 /// by the terminal running broot.
 ///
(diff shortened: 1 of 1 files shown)


Canop:
Please explain the logics of your PRs.

You changed a lot of lines, with apparent goal just being to implement "A missing version is assumed to be a recent, supporting build". Right ? Why ?

chinhkrb113:
Yes, that's the core change. The original logic treated a missing or unparseable version string as "no support," which forced Wezterm into a degraded rendering path even though current builds fully support the Kitty protocol. Since Wezterm doesn't always report a version we can parse, assuming a recent supporting build avoids the false negative that caused #509.

Happy to trim the diff if the extra changes are mostly incidental refactoring — let me know which parts you'd like me to scope back so the PR stays focused on that one behavioral fix.

Canop:
That looks right. Wezterm users are probably using a recent version anyway. Let's try.

PR #1175, opened 2026-06-24
Add Sixel image-preview support

jamison-wilde:
## Add Sixel image-preview support

Adds the Sixel graphics protocol alongside Kitty, so image previews render in Windows Terminal 1.22+ and Sixel-capable Unix terminals (foot, mlterm, xterm). Kitty stays preferred; terminals with neither use the existing Unicode fallback.

- **Commit 1 (`refactor:`)** — extract a protocol-agnostic `graphics` layer
  (`GraphicsRenderer` trait + `GraphicsManager`) out of `kitty`. No behavior change.
- **Commit 2 (`feat:`)** — Sixel renderer via `icy_sixel`, auto-detected (DA1
  query; `CSI 16 t` for cell size on Windows). Probing runs once at startup before
  the input reader thread to avoid racing it. `BROOT_GRAPHICS_PROTOCOL` overrides.

**Dependency:** needs the Windows backend added to `xterm-query` in [this PR](https://github.com/Canop/xterm-query/pull/11) . The `[patch.crates-io]` xterm-query line in the Cargo.toml is left commented per convention; the Windows build compiles once that release publishes, or it is edited manually to remove the comment and use a local copy. 

Tested in Windows Terminal and with Ubuntu in xterm.

image


src/graphics/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// Choose a graphics renderer for this terminal, honoring the
+/// BROOT_GRAPHICS_PROTOCOL override (kitty | sixel | none).
+fn select_renderer(con: &AppContext) -> Option<Box<dyn GraphicsRenderer>> {
+    use crate::kitty::KittyGraphicsDisplay;
+
+    // The legacy `kitty-graphics-display = "none"` config predates Sixel and is
+    // treated as the global "disable all terminal graphics" switch (both Kitty
+    // and Sixel), for backward compatibility. The option name is now a misnomer.
+    if con.kitty_graphics_display == KittyGraphicsDisplay::None {
```
Canop: I don't like that `kitty-graphics-display = "none"` disables all high-res graphics rendering, even if it may temporarily look like it's more backward compatible.

I suggest
- we don't change the behavior but stop documenting the "None" variant
- we make  `kitty-graphics-display = "none"`  only disable Kitty graphics
- we add a new GraphicsDisplay enum, with variants None, Auto (i.e. detect),  Kitty, Sixel

Canop:
Testing it (on Ubuntu with `xterm -ti vt340 -xrm "XTerm*numColorRegisters: 256"`).

It works for some pictures but not all (seems to fail on big ones). Did you observe failures on your own. Did you look into them ?

jamison-wilde:
> Testing it (on Ubuntu with `xterm -ti vt340 -xrm "XTerm*numColorRegisters: 256"`).
> 
> It works for some pictures but not all (seems to fail on big ones). Did you observe failures on your own. Did you look into them ?

I will take a look. I did check on Windows with some large images (4k-ish)and they worked fine and were relatively fast. But it seems to crop the large image weirdly on Ubuntu.

image

It did eventually load (on Ubuntu VM and Windows) a super large IMO scanned image, though took a while.
image

Will dig deeper and see how these can be resolved and what limitations already existed with the kitty/halfblock processing times.  Seems like the slowness is partially in zune_jpeg:

12:49:43.691 [DEBUG] zune_jpeg::color_convert: Using AVX optimised color conversion functions
...
12:50:21.209 [DEBUG] broot::image::image_view: decode image on "...large_scanned_test_image.jpg" **took 37.5383724s**
12:50:21.209 [DEBUG] broot::app::app: cmd_result: CmdResult::NewPanel { purpose: Preview, direction: Right, .. }
12:50:21.211 [DEBUG] broot::image::image_view: image_view must be cleared
???? assume some sixel magic is happening here, but TBD ???? 
12:50:48.730 [DEBUG] broot::graphics::image_data: using rgb
12:50:49.630 [DEBUG] broot::sixel::sixel_renderer: printed sixel image 463x760 into Area { left: 92, top: 1, width: 47, height: 38 }
12:50:49.630 [DEBUG] broot::preview::preview: iv.display(w, disc, area) **took 28.4187583s**

We could always try to add (configurable) limits around image sizes to load in theory as well, if it was systemic regardless of renderer. Scope creeping though. /sweating

jamison-wilde:
Ok, so the image rendering slowness seems to be outside the scope of the sixel changes:

14:10:48.082 [DEBUG] zune_jpeg::color_convert: Using AVX optimised color conversion functions
<< GAP 1 is decoding jpeg
14:11:25.008 [DEBUG] broot::image::image_view: decode image on "...large_scanned_test_image.jpg" took 36.9515265s
14:11:25.010 [DEBUG] broot::image::image_view: image_view try_print_image
14:11:25.010 [DEBUG] broot::graphics: try_print_image area_width: 940, area_height: 760
<<< GAP 2 is 'fitting' image data 
14:11:51.490 [DEBUG] broot::graphics: try_print_image fitted image
vvvv sixel stuff starts vvvvv
14:11:51.490 [DEBUG] broot::sixel::sixel_renderer: sixel print fill_bg
14:11:51.491 [DEBUG] broot::sixel::sixel_renderer: sixel ImageData::from pre
14:11:51.491 [DEBUG] broot::graphics::image_data: using rgb
14:11:51.491 [DEBUG] broot::sixel::sixel_renderer: sixel ImageData::from post
14:11:51.516 [DEBUG] broot::sixel::sixel_renderer: sixel print pre
14:11:52.348 [DEBUG] broot::sixel::sixel_renderer: sixel print post
14:11:52.366 [DEBUG] broot::sixel::sixel_renderer: printed sixel image 463x760 into Area { left: 92, top: 1, width: 47, height: 38 }
14:11:52.367 [DEBUG] broot::graphics: try_print_image drawn: None
14:11:52.367 [DEBUG] broot::image::image_view: image_view try_print_image after
14:11:52.367 [DEBUG] broot::preview::preview: iv.display(w, disc, area) took 27.3564616s

I have some experience with icy_sixel with ratatui-image, so I was pretty sure the sixel perf itself wasn't a problem, but confirmed. 

(Aside: definitely something that interests me to improve, along with maybe asyncing the preview to unblock the UI around previews, but no idea what is involved there yet).

jamison-wilde:
Ok, so the xterm cropping is expected due to _maxGraphicsSize_ defaults of 1000x1000 in xterm. Adding a detection for this OSC geometry value and it now scales properly. _maxGraphicsSize_ can be set to  auto in the ~/.Xresources apparently to override this behavior. 

image

(Not committed yet...working on it)

gustavo-hms:
Hello! I've tested it on Konsole 26.04.2 (which has support for Sixel protocol but not for Kitty's) and it works, but with some issues.

Just for control, this is broot version 1.56.2 on Konsole:

Captura_de_tela_20260625_175527

And that is the result using broot from this PR:

Captura_de_tela_20260625_175621

That horizontal orange bar at the bottom is an artefact and is not present in the original image.

Konsole, as many modern terminal emulators, has support for text reflow on window resizing. So I tested resizing the terminal window and that's the result:
 
Captura_de_tela_20260625_175654

That artifact is not erased as I preview other images:

Captura_de_tela_20260625_175719

jamison-wilde:
> Hello! I've tested it on Konsole 26.04.2 (which has support for Sixel protocol but not for Kitty's) and it works, but with some issues.

Thanks for the feedback! So there is definitely a quirk in sixel because it renders in a block height and images that don't cleanly fit in that will have extra pixels drawn to fill the block...tentatively guessing that may be the issue and the orange is either from another image or some other artifact, but TBD and I could be wrong. Will test it out.

Will check out the other issues, guessing the resize one has something to do with the 'clear this area' calculations using the new area instead of the previous area, or resize not triggering a full TUI re-render in general (which is probably the typical scenario).

Thanks again.

jamison-wilde:
Also confirmed the slowness is not sixel related and is not an issue: **cargo run --release** performs fine for all of the graphics modes....my giant scanned image 10000x16000 jpg takes  ~2secs on kitty and sixel from ctrl left arrow to render.

@Canop Can you confirm if --release profile builds perform well/better for you? If there are still images causing problems for you, LMK, and give me some more details or share the images/deeper config info for your Ubuntu instance? Thanks!

Also, I have merged the scaling fix, though I have left my debug! lines for timing testing for the moment, will remove them.

I will take a look at the Konsole feedback next.

Canop:
> Can you confirm if --release profile builds perform well/better for you?

It's slow, but not so much of a problem: it's usable enough.

Tell me when you think the code is ready and then I'll start reviewing the code, which won't be fast because I don't merge code without really checking it and that's a lot of changes.

jamison-wilde:
@Canop Sure no rush, understood on the timing. I am seeing similar times between kitty, sixel, and halfblock, so I think it is running up against the perf of the image decoding, and resizing pre-rendering...left one debug! line in my PR that makes that clearer.

@gustavo-hms 
The fixes for Konsole are also a good chunk of code as well. Long and short of it there, Konsole sixel is an artifact generating mess, where there are separate sixel and text layers and drawing to one doesn't do anything to the other so most of the easier 'clear this portion of the screen' options don't work. The workaround is to be a bit more 'redraw everything' if we are resizing the whole terminal window, or when hiding the preview image or when switching from image preview to another preview type (directory 'preview', etc. were messy).  I targeted the scope to Konsole were it made sense to not add the extra cost of refreshing to other terms that don't need it, but that may be a point of reviewer discussion/feeback.

@Canop I'll give Gustavo a chance to respond before saying it is ready for your personal review. Thanks!

gustavo-hms:
Hi, @jamison-wilde ! Thanks for your work.

The resizing issue is gone. The image blinks while the window is being resized (certainly due to the redraw) but at the end the resizing artefact is no more.

The horizontal bottom bar remains though. But I consider that a minor issue and I'm OK with it.

jamison-wilde:
@gustavo-hms  So this one is actually what I thought of originally...the sixel 6px 'block' height causing issues, so I actually repro'd and fixed something different before. Working on this now, but it is a bit tricky on the edge cases...basically need to pad images with height%6  to prevent this (and only for Konsole so far /sigh).

jamison-wilde:
@gustavo-hms Ok, so the Konsole sixel banding artifacts redux seems to be resolved. LMK if you see any issues if you get a chance, and appreciate your help!

gustavo-hms:
Almost there!

There's still an horizontal bar at the bottom. It only changed to be more unobstrusive.

Screenshot from 2026-06-27 17-18-55 copia

Screenshot from 2026-06-27 17-23-24

But I must say I'm satisfied with the result. It doesn't need to be perfect, specially if this artifact appears only on Konsole,  as it seems to be the case.

Canop:
@jamison-wilde  You should now update the xterm-query dependency to 0.6

jamison-wilde:
@gustavo-hms yea, that is the sixel pre-filling that, where the the image only needs a couple of pixels of the sixel block height. Used the default bg, but will see what the preview panel is using and how accessible that is.

jamison-wilde:
@Canop Updated the dep for broot, though there is still a dep in terminal-light, that doesn't conflict, just FYI on why there are two xterm-query's in the lock file. 

@gustavo-hms So the fix for the band would be a bit heavy handed. Ultimately the issue is that sixel is 6px chunks, and cells are not typically...the way Konsole applies sixel to the terminal, we would need to change the sixel to a shared multiple of 6px and the cell height, resizing the image just to get rid of this band, and creating some weird edge cases if the font is a prime number for example. (I actually ran into this on another project, a Rust sixel web browser frontend I am working on, where I wanted to cache rendering when scrolling, and had to, at least on Windows Terminal, lock scrolling increments to 60px - 6*10 since on Windows Terminal the 'virtual cells' are always 10px.) 

There is a related Konsole bug around this which is wontfix and ratatui-image throws their hands up at Konsole lol:
https://bugs.kde.org/show_bug.cgi?id=456354

I think the cure is worse some ways. It looks fine, but just feels wrong to have to do that image scaling for this, especially on small images. But I have the simple but hammery fix if anyone has a strong opinion. 

I am taking one last stab, digging into it to see if I can resolve that Konsole issue.  Will keep you posted.

src/display/cell_size.rs
```diff
@@ -43,11 +43,73 @@ pub fn cell_size_in_pixels() -> std::io::Result<(u32, u32)> {
...
+}
+
+/// Parse a `CSI 16 t` reply of the form `ESC [ 6 ; height ; width t`
+/// into `(width, height)` in pixels. Returns `None` if it isn't a
+/// well-formed cell-size (kind `6`) report.
+///
+/// Only used by the Windows `cell_size_in_pixels` (Unix uses `ioctl`), so it's
+/// compiled on Windows and in test builds to avoid a dead-code warning on Unix.
+#[cfg(any(windows, test))]
+pub(crate) fn parse_cell_size(response: &str) -> Option<(u32, u32)> {
```
j4james: Just FYI, ioctl isn't actually guaranteed to return pixel dimensions on Linux. Just as one example, a Linux terminal remoting into another Linux host over a telnet connection would not be able to propagate their pixel dimensions, even though both parties support ioctl. So for best results you should ideally be falling back to a `CSI 16 t` or `CSI 14 t` query when ioctl fails (i.e. when returning bogus pixel values).

jamison-wilde:
Sorry, had unexpected travel, will be back on this.

Canop:
> Sorry, had unexpected travel, will be back on this.

No worry, we're not paid for that

src/display/cell_size.rs
jamison-wilde: Fixed the ioctl issue in the latest changes. Thanks for the heads up!

Canop:
Tell me when it's ready, and don't worry for the conflicts, I'll handle them.

jamison-wilde:
> Tell me when it's ready, and don't worry for the conflicts, I'll handle them.

I had 'Fable' do a comprehensive review on the PR changes, working through that to see if there is anything real there.

jamison-wilde:
Ok, added a bunch of tests and fixed a few issues. I think it is ready to go. I can do the conflict rez if you want, just let me know. Thanks for your patience!

Canop:
Good. I'll review then probably merge it (from my previous tests and partial review it looked good).

It doesn't come at the ideal time (I started a new job last week and will go to holidays) so there may be some delay but it won't be forgotten.

Testers: your feedback is crucial.

gustavo-hms:
It looks good to me.

Canop:
Taking it from now, merging it with some changes.

Canop:
It's now in main. Thanks all for the code and test. I hope I didn't broke it with my further refinements...

I'l try to release a new version of broot next week-end.

Canop:
broot 1.59.0 released

If you compile broot yourself, you have to add the feature, eg `cargo install broot --features "clipboard sixel" --locked`

PR #1186, opened 2026-07-26
Quote paths containing a backslash

Cyrus580529:
`path_to_string` only quotes when the path holds whitespace or a quote. A Windows path like `C:\Users\me\dev` has neither, so it reaches the outcmd file unquoted, and the `eval` in the `br` launcher strips the backslashes before `cd` sees them.

Putting the backslash in that character class gives it the same single quoting spaces already get. On Windows 11 with Git Bash the outcmd line goes from `cd C:\Users\me\dev` to `cd 'C:\Users\me\dev'`, and `:cd` works again.

The test checks `path_to_string` with the String target on three inputs: nothing special, a space, a backslash.

`test_path_normalization` fails on Windows on main, before and after this change.

### DIFF
--- a/src/verb/execution_builder.rs
+++ b/src/verb/execution_builder.rs
@@ -483,7 +483,7 @@ impl<'b> ExecutionBuilder<'b> {
         }
         // first we replace single quotes by `'"'"'` (close the single quote, add an escaped
         // single quote, and reopen the single quote)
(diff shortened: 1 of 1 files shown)


Canop:
@Cyrus580529 Can you address Copilot's remarks ? The first one at least.

Cyrus580529:
Fixed, added a test for the single-quote case too.

PR #1192, opened 2026-08-08
fix: correct highlight offset when a matched name is truncated on the left

YuriNachos:
When a matched file/dir name is longer than its column and `MatchedString::cut_left_to_fit` removes `removed_char_count` (rcc) characters from the LEFT, the surviving match positions must reindex by rcc. `NameMatch::cut_after(k)` returns positions `p >= k` reindexed as `p - k`, so the correct argument is `cut_after(rcc)`. The code used `cut_after(rcc - 1)`, which shifts every surviving highlight one character to the RIGHT of where it belongs — and when the last removed char was itself matched (`p == rcc - 1`) it also keeps a phantom highlight on the first visible character. Reachable via the tree name column (`displayable_tree.rs`) and the trash view (`trash_state.rs`) whenever a matched name is truncated on the left. Shipped since the root commit.

## Fix
One character: `nm.cut_after(removed_char_count - 1)` → `nm.cut_after(removed_char_count)`.

## Test plan
`cargo test` — green, including a new in-module `cut_left_to_fit_reindexes_match_positions` test (`#[cfg(test)]`, since `MatchedString` is private) that asserts positions reindex to `[0,1]` (today `cut_after(1)` yields `[1,2]`) and a phantom-highlight case. CHANGELOG bullet added.

### DIFF
--- a/src/display/matched_string.rs
+++ b/src/display/matched_string.rs
@@ -109,7 +109,7 @@ impl<'a> MatchedString<'a> {
             self.name_match = self
                 .name_match
                 .take()
-                .map(|mut nm| nm.cut_after(removed_char_count - 1));
+                .map(|mut nm| nm.cut_after(removed_char_count));
         }
         removed_char_count
     }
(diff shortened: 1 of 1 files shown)


YuriNachos:
Now that #1190 and #1191 are in, this is the last of the three small fixes I had queued in the same area — here the highlight positions shift one character right whenever cut_left_to_fit truncates a matched name on the left, because the reindex passes cut_after(rcc - 1) instead of cut_after(rcc). Does the one-argument fix look right to you, or would you like a visible test case (long matched name in a narrow column) before it goes in? Either way I'm happy to add whatever makes it easiest to verify.

Canop:
I'll have a look as soon as I find enough time to try reproduce it.

Canop:
I'm sorry but I don't manage to reproduce it. Maybe I don't understand the problem. Can you post a screenshot exhibit?

YuriNachos:
Here is a reproducer — the key detail I should have given earlier: it takes a **path search** (`p/…`), not a name search. `cut_left_to_fit` is only reached from the subpath branch of `write_line_label`, and the path has to be long enough to be truncated on the left.

```bash
mkdir -p "somewhere/warawara-aaaaaaaaaaaaaaaaaaaaaaaaaaaa/target-bbbbbbbbbb"
touch "somewhere/warawara-aaaaaaaaaaaaaaaaaaaaaaaaaaaa/target-bbbbbbbbbb/f.txt"
broot somewhere        # in a 44-column terminal
```

then type `p/target` in the search box.

On current main (dfd25e0c) the last tree line reads `…aaaaaaa/target-bbbbbbbbbb/f.txt` and the match highlight paints **`arget-`** — it starts one character late, missing the leading `t` and painting the following `-`. With the PR branch the same line paints **`target`** exactly:

left-cut highlight, main vs PR

That's the off-by-one: the visible part of the path starts at `removed_char_count`, so re-indexing the kept match positions against `removed_char_count - 1` shifts them all one to the right.

Canop:
Thanks for this very detailed reproducer

PR #1199, opened 2026-08-14
feat: add the :no_action internal

### LINKED ISSUE #328: :no_action verb, to disable keys
This verb doing nothing would be used to disable a key in a specific mode or for  a specific kind of selection.

Initially requested in https://github.com/Canop/broot/issues/185#issuecomment-765803736

VXNCXNX:
Fix #328.

Adds the `:no_action` internal you asked for:

```hjson
{
    key: ctrl-t
    internal: ":no_action"
    apply_to: "directory"
}
```

## Why this isn't just `unbind`

broot already lets you unbind a key by declaring a verb with keys and no execution (`verb_store.rs:514-520` → `unbind_key`). That strips the key from every verb, everywhere.

`:no_action` is different because it's a *verb*, so it goes through `find_key_verb`, which filters on `panels`, `selection_condition` and `accepts_extension` before dispatching — and its own comment notes that "there can be several verbs with the same key and not all of them can apply". That's what makes the conditional case in your issue possible: disabling a key *"in a specific mode or for a specific kind of selection"*, which unbinding can't express.

The shadowing works because `VerbStore::new` pushes conf verbs before `add_builtin_verbs()`, and `find_key_verb` returns the first match.

## Implementation

Three lines: the `Internals!` entry (`need_path: false`), the doc row in `conf_verbs.md`, and a CHANGELOG entry.

No dispatch arm is needed — `on_internal_generic` already ends in `_ => CmdResult::Keep`, and `no_action` isn't input-related, so it lands there naturally. I did write an explicit `Internal::no_action => CmdResult::Keep` arm first and then removed it: it was dead code, no other internal has one, and the crate still builds without it.

## Two decisions worth your call

**Not registered as a builtin verb.** I left `:no_action` out of `add_builtin_verbs`, so it's usable from conf but not typable in the input. That follows the precedent of `next_dir`, `next_same_depth` and `line_up_no_cycle`, which are documented in `conf_verbs.md` without being registered. It also avoids making `:no` ambiguous against the registered `:no_sort`. One line to change if you'd rather have it typable.

**Input mode.** Since `is_input_related()` is false for it, a printable key bound to `:no_action` still types its character in Input mode; the disabling applies in Command mode. That seemed right, but it's a choice rather than an obvious consequence.

## Verification

`cargo build` clean, `cargo test` 54 + 7 passed, `cargo clippy --all-targets` adds no warnings for the touched file (the one hit at `internal.rs:217` is pre-existing — I confirmed it against a stashed baseline).

End to end, with `BROOT_CONFIG_DIR` pointed at a scratch conf:

```
internal = ":no_action"       -> conf loads, broot starts
internal = ":no_action_typo"  -> Invalid verb configuration: not a known internal: no_action_typo
```

I couldn't drive an actual keypress in this headless environment, so the suppression itself is verified by reading `find_key_verb` rather than by pressing ctrl-t.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### next
+- new `:no_action` internal, doing nothing, which can be used to disable a key - Fix #328
 - High-Res images in Rio terminal (detect it to enable the Kitty image protocol) - Fix #1179
 - fix content-exact match line number off-by-one when the match starts at the first byte of a line (broot jumped to the line above) - Thanks @YuriNachos
 - fix: detect a duplicate broot server name instead of silently overtaking the running server - Fix #1065 - Thanks @YuriNachos
--- a/src/verb/internal.rs
+++ b/src/verb/internal.rs
@@ -90,6 +90,7 @@ Internals! {
     next_dir: "select the next directory" false,
     next_match: "select the next match" false,
     next_same_depth: "select the next file at the same depth" false,
+    no_action: "do nothing (can be used to disable a key)" false,
     no_sort: "don't sort" false,
     open_leave: "open file or directory according to OS (quit broot)" true,
     open_preview: "open the preview panel" true,
--- a/src/verb/verb_store.rs
+++ b/src/verb/verb_store.rs
@@ -264,6 +264,7 @@ impl VerbStore {
             .with_key(key!(shift - backtab))
             .with_key(key!(backtab));
         self.add_internal(next_match).with_key(key!(tab));
+        self.add_internal(no_action);
         self.add_internal(no_sort).with_shortcut("ns");
         self.add_internal(open_stay)
             .with_key(key!(enter))
--- a/website/src/conf_verbs.md
+++ b/website/src/conf_verbs.md
@@ -483,6 +483,7 @@ invocation | default key | default shortcut | behavior / details
 :next_dir | - | - | select the next directory
 :next_match | <kbd>tab</kbd> | - | select the next matching file, or matching verb or path in auto-completion
 :next_same_depth | - | - | select the next file at the same depth
+:no_action | - | - | do nothing (can be used to disable a key)
 :no_sort | - | ns | remove all sorts
 :open_leave | <kbd>alt</kbd><kbd>enter</kbd> | - | open the selected file in the default OS opener and leave broot
 :open_preview | - | - | open the preview panel


Canop:
> I couldn't drive an actual keypress in this headless environment, so the suppression itself is verified by

Please stop pushing unchecked LLM authored PRs. If you're interested in a problem, at least test. If you're not interested in an issue, don't bother.

VXNCXNX:
Fair, and I'll close this.

Two things you're right about. The issue is from 2021 and I didn't check whether you still wanted it, so this arrived as an unasked feature rather than a fix. And I shipped the key behaviour, the actual key suppression, verified by reading `find_key_verb` instead of by running it. I said so in the description, but saying so isn't the same as testing it, and it shouldn't have gone out that way.

Sorry for the noise.

Canop:
Actually it's probably useful. And now that we both spent time on this, and that I got the issue remembered, I'll probably merge and test or do it another way. So I'll reopen it.

Just please, don't do that again, it's extremely tiring to read an issue, then a whole PR, and just notice at the end that the author didn't test and probably didn't care.

VXNCXNX:
Fair, and you were right. I rebased on main and then actually ran it, and the PR as you saw it did not work at all:

```
$ broot -c ':no_action;:print_path'
No verb matches "no_action"
```

Declaring the internal in the `Internals!` macro is not enough, it also has to be registered in `verb_store.rs`, which I had missed. So the branch you were looking at added a documented verb that could not be invoked. That is exactly the thing you were complaining about, and I would have caught it by running it once.

One line added:

```rust
self.add_internal(no_action);
self.add_internal(no_sort).with_shortcut("ns");
```

Now verified in a real terminal, via tmux against a tree of `f1`, `f2`, `sub`. With `down` bound to `:no_action`:

```
verbs: [ { key: down, internal: ":no_action" } ]
```

```
after 3x Down:   no selection marker moved
then 1x Up:      ▶└──sub      (unbound keys still work)
```

and with `down` left alone, the same three presses walk the tree normally. So the key is inert rather than the app being wedged.

The CHANGELOG conflict was just the reworded Rio line, I kept yours.

`cargo test` is 61 passed, `cargo clippy --all-targets` is clean. `cargo fmt --check` flags `verb_store.rs`, but it flags the same three hunks with my change stashed, so it is pre-existing and I left it alone.

*Disclosure: written with AI assistance (Claude Code). The point above stands regardless: I should have run it before opening it the first time, and this time I did.*

Issue #1145, opened 2026-03-22
Allow multiple commands in external verb


xbwwj:
Is it possible to eval multiple command in external verb? For example, execute the following command:

```bash
mkdir foo && touch foo/bar
```

```json
// This config is not valid in current version
 {
    "invocation": "foobar"
    "external": "mkdir foo && touch foo/bar"
    "leave_broot": false
 }
```

### Current workaround

This can be worked around with `bash -c`, but the syntax looks quite messy.

```json
{ "external": "bash -c \"mkdir foo && touch foo/bar\"" }
```

### Possible syntaxs

1. array
    ```json
    { "external": [ "mkdir foo", "touch foor/bar" ] }
    ```
2. semicolon (in accord with `cmd`)
    ```json
    { "external": "mkdir foo; touch foor/bar" }
    ```
 
### Related issues

- <https://github.com/Canop/broot/issues/933> is similar, but only verb sequence is supported.

Canop:
In case it works on the shell, you now can use shell_command:

    { shell_command: "mkdir foo && touch foo/bar" }

Does this solve the issue for you ?

xbwwj:
Sorry for the delay. This does solve the issue. Thanks very much.

PR #1204, opened 2026-08-18
Add a shell attribute to run external verbs through a shell (#1145)

ChrisJr404:
Adds a `shell` attribute for external verbs so a command can chain with `&&`, `;` or pipes without leaving broot, closing the gap in #1145.

Right now an `external` is tokenized and run as a single program via `Command::new`, so `mkdir foo && touch foo/bar` just passes `&&` as an argument to `mkdir`. The docs already show `&&` working, but only with `from_shell: true`, which needs the `br` function and quits broot. There was no way to chain when you want `leave_broot: false`.

With `shell: true` broot builds the command line the same way it does for the parent-shell case (placeholders filled, paths quoted) and hands it to `sh -c` on unix or `cmd /C` on Windows. Env vars are left for the shell to expand. Without the flag nothing changes, so existing verbs run byte-for-byte as before.

Example:

```hjson
{
    invocation: "mkfile {name}"
    external: "mkdir -p {directory}/foo && touch {directory}/foo/{name}"
    shell: true
    leave_broot: false
}
```

I added a `Launchable::shell_program` constructor, a `shell` field on `ExternalExecution` (default false), wired it through `VerbConf`/`verb_store`, and branched the three exec paths (leave-broot, stay-in-broot single, and per-selection/merged multi-selection). Docs get a new table row plus a "Chaining commands" section, and there's a changelog entry.

Executing through a real shell is hard to unit-test, so the added test covers the parse/build path — it asserts that a command with `&&` survives tokenizing and re-joining into the exact string that gets handed to `sh -c`. `cargo test` is green (147 passed). Formatting matches the tree (checked with nightly rustfmt, which the repo's `rustfmt.toml` targets) and clippy shows no new warnings on the touched code.

I picked an explicit opt-in over auto-detecting operators or the array/semicolon syntaxes floated in the issue, to keep it backward-compatible and predictable, but happy to switch to whatever spelling you prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### next
+- new `shell_command` verb attribute: run a command through a shell (`sh -c` / `cmd /C`) so `&&`, `;` and pipes work, without leaving broot - Fix #1145
 - fix invalid official Mac binary (duplicate linked dylib) with new build chain - Fix #1194
 - Sixel graphics support for image preview, auto-detected: works in iterm2, Windows Terminal 1.22+ and Sixel-capable Unix terminals (foot, mlterm, xterm built with Sixel, recent WezTerm). Kitty remains the preferred protocol when available - Fix #568 - Thanks @jamison-wilde
 - High-Res images in Rio terminal (detect it to enable the Kitty image protocol) - Fix #1179
--- a/src/app/panel_state.rs
+++ b/src/app/panel_state.rs
@@ -834,7 +834,10 @@ pub trait PanelState {
                 cc,
             ),
             VerbExecution::External(external) => {
-                self.execute_external(w, verb, external, invocation, app_state, cc)
+                self.execute_external(w, verb, external, false, invocation, app_state, cc)
+            }
+            VerbExecution::ShellCommand(external) => {
+                self.execute_external(w, verb, external, true, invocation, app_state, cc)
             }
             VerbExecution::Sequence(seq_ex) => {
                 self.execute_sequence(w, verb, seq_ex, invocation, app_state, cc)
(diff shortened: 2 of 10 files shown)


Canop:
I approve the general idea.

There are a few things which should IMO be different regarding internal design and config.

As it's a different kind of execution, validated and processed in different ways, I'd rather have a new variant in `VerbExecution`, eg `ShellCommand`.

Regarding the configuration, and the `VerbConf` struct, I'd prefer a new parameter, so it would be configured as

```
{
    invocation: "mkfile {name}"
    shell_command: "mkdir -p {directory}/foo && touch {directory}/foo/{name}"
    leave_broot: false
}
```

Your opinion welcome.

ChrisJr404:
Thanks, that framing is much cleaner - I've re-architected it to match.

There's now a dedicated `VerbExecution::ShellCommand` variant instead of a flag on `ExternalExecution`, and a `shell_command` parameter on `VerbConf` (the old `shell` boolean is gone). A verb configured like your example:

```hjson
{
    invocation: "mkfile {name}"
    shell_command: "mkdir -p {directory}/foo && touch {directory}/foo/{name}"
    leave_broot: false
}
```

now parses into `VerbExecution::ShellCommand(...)` and is dispatched separately from a plain `external`, running the line through `sh -c` (`cmd /C` on Windows). Placeholders are still expanded and paths quoted before it reaches the shell, so env vars are left for the shell to resolve. Docs and changelog updated to the new spelling.

`cargo test` is green (147 passed), formatting matches the tree, and clippy shows no new warnings. Let me know if you'd like any further tweaks.

Canop:
At this point, the `from_shell` attribute, which was never very clear, becomes very confusing.

So it looks like the right occasion to also rename it, eg to `run_in_parent_shell` (with an alias for compatibility).

Either you do it or I'll do it when merging your PR (I have several pages of the doc to change anyway).

Canop:
Taking it from there to fix from_shell and edit the doc.

PR #1207, opened 2026-08-28
Replace libgit2 with gix, making broot pure rust

Canop:
For a very long time, I've waited for gix to be ready to replace ligbgit2.
It was now possible, so now, **broot is pure rust!**

- git status now computed with gix (pure Rust) instead of libgit2: no more C dependency, faster on big repositories
- in detached HEAD state, the status line shows the short commit id instead of `HEAD`
- git status: staged files are shown (`A` added, `S` staged modification, `R` renamed); new `git_status_staged` skin entry
- git status of a subdirectory of a big repository is computed on that subdirectory only, which is much faster
- git status: an untracked directory gets the `N` mark too
- the status line shows how many commits the branch is ahead (`↑`) or behind (`↓`) its upstream
- `{git-root}` and `{git-name}` now work when the selection is a file

Note: tests on Windows would be welcome, I'm unsure regarding handling of crlf.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 - git status: an untracked directory gets the `N` mark too
 - the status line shows how many commits the branch is ahead (`↑`) or behind (`↓`) its upstream
 - `{git-root}` and `{git-name}` now work when the selection is a file
+- alt-g toggles the git status filter (`:toggle_git_status`)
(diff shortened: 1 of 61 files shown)


Canop:
With diff preview, and the same "panel_right for details", "tab for next chunk" logic than for search matches: 

image

PR #1218, opened 2026-09-19
Bump chacha20 from yanked 0.10.1 to 0.10.2

muzimuzhi:
See https://crates.io/crates/chacha20/0.10.1.

From https://github.com/RustCrypto/stream-ciphers/commit/dc302eca24cb5a93ebe465578bb8b4d90948b07f, 

> The yanked versions contain UB in SSE2 backend


Canop:
There was no possible impact of the bug on broot but this avoids pulling a yanked crate so avoids useless reports, so OK ^^

